PR #17, opened 2016-04-01
Adding an interval check, if you don't want to run every time

spuranam:


chris-rock:
Thanks @spuranam for adding this PR. I like the addition. Could you do me one favor:
- fix the issues highlighted in https://travis-ci.org/chef-cookbooks/audit/builds/120182663
- add a short description in this PR, why you need that

andy-dufour:
thanks @chris-rock .. I think @spuranam accidently hit the new pull request button, I want to scope this delay interval per compliance profile and then I'll push some new changes up and ping for a review in the next couple of days ;)

andy-dufour:
@chris-rock @spuranam @srenatus 

This should be ready for review now. Let me know if you have any comments.

andy-dufour:
Also - the reason this is important is if you have the audit cookbook in your nodes runlist.

There are customers that have inspec/compliance profiles that take 10+ minutes to run, I know one implemented thousands of lines of Inspec for DISA STIG on Windows that takes 13+ mins. This is a significant resource drain, and if your chef-client is executing frequently (say 30 mins or less) you're spending a lot of time auditing. 

By making this audit interval say, once a day, a customer can restrict it from running all the time. By making this per profile we ensure that when a new profile is added it doesn't need to wait a maximum of <interval> to execute for the first time.

chris-rock:
@andy-dufour This is a great addition. It starts to catch the use where the audit schedule is different from the converge run. Customers also mentioned requests like: "I want to run my audit every sunday morning". Therefore I am really happy to support the use case, but I am not sure yet about the interface. I think we need some time to learn more about it.

To enable the usecase quickly and give customers a way to use it, I would like to do the following:

- make an extra recipe to cover that use case (eg. `interval.rb`) instead of adapting the `default.rb`
- all code for interval handling should not be part of the core libraries (eg. https://github.com/chef-cookbooks/audit/pull/17/files#diff-fa67e9573e0ece6499cc04e556ec4d55R107), maybe we just write a interval library?
- I would like to mark this feature as experimental to reserve the right to adapt and improve the interface at all times

srenatus:
https://www.youtube.com/watch?v=RamqbWBouMM :+1:

andy-dufour:
@chris-rock how do you feel about this? (Once tests pass, of course)

chris-rock:
@andy-dufour That looks great. One minor thing I was thinking about: Should we wrap the `not_if ` around an `include_recipe 'audit::default'`? What do you think?

andy-dufour:
I like it for DRY reasons and it would be cleaner (plus if the default recipe changes we would pull in the upstream fixes).

The use case I'm trying to solve for and the reason I didn't do that is that I'm doing an interval per each individual compliance profile -- so if a user has a 'once per week' interval, and they add a profile in mid-week, they don't need to wait days for that profile to run the first time.

Perhaps we should simplify this though, and say it's a blanket 'all your profiles will run once per interval', until we flesh out the interface and bake it in a little more.

chris-rock:
@andy-dufour sounds like a good approach. Lets try to harmonize this in future PRs. 

I like to thank @spuranam and @andy-dufour for this great addition.

PR #19, opened 2016-04-04
Use move to avoid cross-device error

alexpop:
File rename won't work across devices, throwing an "Invalid cross-device link" exception.

Thanks Marcus for reporting the issue!


srenatus:
Thanks @alexpop and Marcus! :+1:

PR #20, opened 2016-04-06
work with token and direct compliance server API

srenatus:
This adds the ability to directly talk to a compliance server, without
passing through a chef-server. To ensure proper authentication, this
needs a user (access) token:

    token = '....'

    # iterate over all selected profiles
    node['audit']['profiles'].each do |owner_profile, enabled|
      next unless enabled
      o, p = owner_profile.split('/')

      compliance_profile p do
        owner o
        server URI.parse('http://192.168.33.1:8080/api/')
        token token
        action [:fetch, :execute]
      end
    end

    compliance_report 'chef-server' do
      server URI.parse('http://192.168.33.1:8080/api/')
      token token
      organization 'brewinc'
    end if node['audit']['profiles'].values.any?

There're two big __buts__ here, though:

1. It's expecting an access token; so to be of real use in automation setups, something surrounding this chef-run needs to exchange a refresh token for an access token and use `ENV['ACCESS_TOKEN']` (or similar) to pass it into the recipe.
2. Compliance will only _accept reports_ for users (identified via the token) that are authenticated using an oc_id connector (i.e., coming "from a chef server") -- and thus this doesn't make testing the audit-cookbook <--> compliance interaction that much easier: it still needs a chef server in the mix.


srenatus:
example recipe

```ruby
token = 'eyJhbGciOiJSUzI1NiIsImtpZCI6InVQUExHVjFiVDg4YTNnZ0MtQ1hxUDdfTlgzTGZiblZSeHhYUTQwV2l0cXI0Z0pRTGw3T2........'

# iterate over all selected profiles
node['audit']['profiles'].each do |owner_profile, enabled|
  next unless enabled
  o, p = owner_profile.split('/')

  compliance_profile p do
    owner o
    server URI.parse('http://192.168.33.1:8080/api/') # <---
    token token # <---
    action [:fetch, :execute]
  end
end

# report the results
compliance_report 'chef-server' do
  server URI.parse('http://192.168.33.1:8080/api/')
  token token # <---
  direct true # <---
  owner 'admin' # <---
end if node['audit']['profiles'].values.any?
```

libraries/profile.rb
```diff
@@ -46,20 +47,35 @@ class ComplianceProfile < Chef::Resource
...
+      if token # go direct
+        reqpath ="owners/#{o}/compliance/#{p}/tar"
+        url = construct_url(reqpath, server)
+        puts "URL: #{url}"
+
+        tf = Tempfile.new('foo', Dir.tmpdir, 'wb+')
+        tf.binmode
+        Net::HTTP.start(url.host, url.port) do |http|
+          http.use_ssl = url.scheme == 'https'
+          http.verify_mode = OpenSSL::SSL::VERIFY_NONE # FIXME
```
chris-rock: could we make it an attribute and set the default to false?

libraries/report.rb
```diff
@@ -28,13 +28,32 @@ class ComplianceReport < Chef::Resource
...
-      Chef::Config[:ssl_verify_mode] = :verify_none
+      # resolve owner
+      o = return_or_guess_owner
 
-      url = construct_url(::File.join('/organizations', org, 'inspec'))
-      # Chef::Log.debug "url: #{url}"
-      rest = Chef::ServerAPI.new(url, Chef::Config)
-      rest.post(url, blob)
+      if token
+        if direct
```
chris-rock: instead of direct we may use `:chef_server` & `:chef_compliance` as targets with default to chef server. What do you think?

chris-rock:
@srenatus very good, I like the improvement. Just some minor questions

libraries/profile.rb
srenatus: I'm afraid I don't get it. It's not boolean but a string property that is `nil` if unset and some string if it was provided.
So currently, it's a property of a resource. You can use attributes to set it in your _recipe_ ([see here](https://github.com/chef-cookbooks/audit/pull/20/files#diff-ee5748cee3a2d43791dd1e832193194aR29)) -- I haven't had a need to touch the default recipe here.

Having a resource break its interface and _reach out_ into node attributes doesn't seem right to me...

libraries/report.rb
srenatus: I've changed it to `variant`. Note that this is not what triggers the endpoint, but only the _way it is used in compliance_: should it use a chef-server connector token and setup a proper chef-source for the node + env (`:chef`), or is it merely a non-chef-connector token (`:compliance`) and the stuff should be owned by the owner that is given via the other property.

If it's using the chef-gate API or directly calls out to a compliance API endpoint is determined by the presence of the `token` property.

libraries/profile.rb
chris-rock: @srenatus of course we should not use node attributes directly in a library. I was thinking about making it an option to compliance profile:
```
compliance_profile p do
    owner o
    server URI.parse(ENV['COMPLIANCE_API'])
    insecure true
    token token
    action [:fetch, :execute]
end
```

srenatus:
Thanks @alexpop for finding the nasty bug resolved in bfeb004 :+1:

@chris-rock is this what you had in mind?

srenatus:
Rebased & refactored.

recipes/default.rb
```diff
@@ -17,16 +17,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+token = node['audit']['api_token']
```
chris-rock: Should this be `token = node['audit']['token']` ?

attributes/default.rb
```diff
@@ -16,6 +16,10 @@
 # limitations under the License.
 #
 
+default['audit']['server'] = nil
+default['audit']['api_token'] = nil
```
chris-rock: Isn't this the `default['audit']['token']`?

.kitchen.yml
```diff
@@ -59,7 +59,30 @@ platforms:
...
+        token: <%= ENV['COMPLIANCE_CHEF_TOKEN'] %>
+        profiles: *profiles
+  - name: compliance-pure # compliance api with compliance token
+    run_list:
+      - recipe[audit::default]
+    attributes:
+      audit:
+        variant: compliance
+        server: <%= ENV['COMPLIANCE_API'] %>
+        token: <%= ENV['COMPLIANCE_TOKEN'] %>
```
chris-rock: I would like to use `token: <%= ENV['COMPLIANCE_ACCESSTOKEN'] %>` to make clear, that we are not accepting refresh tokens here

chris-rock:
Awesome @srenatus, I just found some minor nitpicking things :-)

chris-rock:
I merge it, once its green

chris-rock:
Awesome @srenatus

cjohannsen81:
Hey, just learned that there has to be a slash at the end (URI object). Otherwise it will fail.

PR #22, opened 2016-04-15
adding ability to handle offline compliance server

jeremymv2:
While testing I noticed a couple of scenarios:

1. A client with a time/date skew simply getting a "401 Unauthorized" error while attempting to download profiles failing the converge.
2. A client failing to converge if the Compliance server API was unavailable.

These changes give the ability to handle these exceptions and allow the node to still continue with the rest of the converge.


chris-rock:
Awesome addition @jeremymv2

PR #24, opened 2016-04-16
Misc updates

tas50:
Syncing this cookbook up with our other community cookbooks


chris-rock:
Thanks @tas50 for the housekeeping!

Issue #3, opened 2016-03-15
add option to fail chef run, if the audit failed


chris-rock:
We should be able to define a node attribute to tell the chef client if it should fail or pass in an event of a detected compliance issue.

jeremymv2:
The above commit handles this.  Simply set `default['audit']['fail_if_any_audits_failed'] = true`

Result of failed audits and is true value is:

```
  * compliance_report[chef-server] action execute[2016-04-21T19:17:47+00:00] INFO: 44 and true


    ================================================================================
    Error executing action `execute` on resource 'compliance_report[chef-server]'
    ================================================================================

    RuntimeError
    ------------
    44 audits have failed.  Aborting chef-client run.

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/audit/libraries/report.rb:52:in `block (2 levels) in <class:ComplianceReport>'
    /var/chef/cache/cookbooks/audit/libraries/report.rb:24:in `block in <class:ComplianceReport>'

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/audit/recipes/default.rb

     36: compliance_report 'chef-server' if node['audit']['profiles'].values.any?

    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/audit/recipes/default.rb:36:in `from_file'

    compliance_report("chef-server") do
      action [:execute]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      declared_type :compliance_report
      cookbook_name "audit"
      recipe_name "default"
    end

    Platform:
    ---------
    x86_64-linux

[2016-04-21T19:17:47+00:00] INFO: Running queued delayed notifications before re-raising exception

Running handlers:
[2016-04-21T19:17:47+00:00] ERROR: Running exception handlers
Running handlers complete
[2016-04-21T19:17:47+00:00] ERROR: Exception handlers complete
Chef Client failed. 3 resources updated in 02 seconds
[2016-04-21T19:17:47+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2016-04-21T19:17:47+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2016-04-21T19:17:47+00:00] ERROR: compliance_report[chef-server] (audit::default line 36) had an error: RuntimeError: 44 audits have failed.  Aborting chef-client run.
```

chris-rock:
[...]

PR #25, opened 2016-04-18
failing converge if any audits failed

jeremymv2:
Simple implementation of #3 

Fails converge _after_ the report submission if there are any audit failures.


chris-rock:
Thanks @jeremymv2 for this cool addition

PR #27, opened 2016-04-18
Add our github templates

tas50:


chris-rock:
Thank you very much for adding @tas50

PR #29, opened 2016-04-22
update README.md with client version requirement

jeremymv2:
### Description

Chef client < 12.5.1 is not able to use the newer `use_automatic_resource_name` method:
```
NameError
---------
undefined local variable or method `use_automatic_resource_name' for ComplianceProfile:Class
```
This change simply documents the client version required to utilized the new custom resource method.


jeremymv2:
Stalled build resulted in termination, not sure what's up with that.

chris-rock:
Thanks @jeremymv2 for bringing this up. I added #30 , since we should support older versions

chris-rock:
Awesome addition @jeremymv2 I have a look, why travis makes strange errors.

PR #56, opened 2016-05-26
changed access token handling

cjohannsen81:
### Description

Changes handling of the token variable.

### Issues Resolved

token was nil because of some resource handling

### Check List

checked against Windows and Linux


chris-rock:
Awesome fix @cjohannsen81 Thanks for finding that issue

PR #57, opened 2016-05-26
test-kitchen example for Chef Compliance direct reporting

chris-rock:
This PR adds an example for Chef Compliance direct reporting


chris-rock:
Thanks @alexpop for the idea to split up the `.kitchen.yml` into a specific windows and linux file

alexpop:
Thanks Chris for these improvements, merging once green!

PR #59, opened 2016-06-07
http_rescue not required with tempfile

Anirudh-Gupta:
### Issues Resolved
Throwing error at line 40 because its not a http request(when integrating with chef server). No response code.

### DIFF
--- a/libraries/profile.rb
+++ b/libraries/profile.rb
@@ -90,9 +90,7 @@ class ComplianceProfile < Chef::Resource # rubocop:disable Metrics/ClassLength
         Chef::Config[:ssl_verify_mode] = :verify_none # FIXME
 
         rest = Chef::ServerAPI.new(url, Chef::Config)
-        tf = with_http_rescue do
           rest.binmode_streaming_request(url)
-        end
       end
 
       case node['platform']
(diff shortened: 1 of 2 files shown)


mhedgpeth:
I'm having this problem as well.

jejohns:
+1 Thank you!

cheeseplus:
+1

jeremymv2:
I don't believe this should have been merged as is.  The problem is that it short-circuits the error response code messages by no longer using the `with_http_rescue` method, which handles errors in the `handle_http_error_code` method.  Now the error handling will no longer occur.

https://github.com/chef-cookbooks/audit/blob/master/libraries/helper.rb#L40

itmustbejj:
@jeremymv2 As I understand the issue that was addressed, `rest.binmode_streaming_request` returns a Tempfile object, not an HttpResponse object (probably butchering actual class names). So there is no `code` attribute to check on the Tempfile object. I think the with_http_rescue should not have been there in the first place.

itmustbejj:
@jeremymv2 Looks like it was introduced unintentionally when the code was rewritten here: https://github.com/chef-cookbooks/audit/commit/0eeefc51bf8b17be4578bb792bfed6b0cf0a239c

smurawski:
https://github.com/chef-cookbooks/audit/pull/62 re-introduces the `with_http_rescue` since there is an http call in the mix - however, it checks the output object to see if it has a response code before dropping into the specific http response handling code.  This should return the file object unimpeded and maintain the troubleshooting benefits of helpful http errors.

PR #63, opened 2016-06-12
Fix compliance refresh_token

jeremymv2:
### Description

This fixes the refresh_token authentication which wasn't working correctly.
It has been tested with ubuntu 14.04 and windows 2012R2

### Check List

### DIFF
--- a/libraries/helper.rb
+++ b/libraries/helper.rb
@@ -49,7 +49,7 @@ module ComplianceHelpers
   end
 
   # exchanges a refresh token into an access token
-  def retrieve_access_token(server_url, refresh_token, insecure = true)
+  def retrieve_access_token(server_url, refresh_token, insecure)
     _success, _msg, access_token = Compliance::API.post_refresh_token(server_url, refresh_token, insecure)
     # TODO we return always the access token, without proper error handling
     access_token
--- a/libraries/helper.rb
+++ b/libraries/helper.rb
@@ -13,7 +13,7 @@ module ComplianceHelpers
 
   def construct_url(server, path)
     # sanitize inputs
-    server+='/' unless server =~ %r{/\z}
+    server << '/' unless server =~ %r{/\z}
     path.sub!(%r{^/}, '')
     server = URI(server)
     server.path = server.path + path if path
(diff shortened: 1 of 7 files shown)


andy-dufour:
Makes sense and LGTM 👍

libraries/helper.rb
```diff
@@ -47,8 +49,8 @@ def with_http_rescue(&block)
   end
 
   # exchanges a refresh token into an access token
-  def retrieve_access_token(server, refresh_token)
-    _success, _msg, access_token = Compliance::API.post_refresh_token(url, refresh_token, options['insecure'])
+  def retrieve_access_token(server_url, refresh_token, insecure = true)
```
chris-rock: Can we set the default for insecure to false here? I think we should assume secure configurations in our code.

libraries/profile.rb
```diff
@@ -61,7 +62,7 @@ class ComplianceProfile < Chef::Resource # rubocop:disable Metrics/ClassLength
 
       # retrieve access token if a refresh token is set
       access_token = token
-      access_token = retrieve_access_token unless refresh_token.nil?
+      access_token = retrieve_access_token(server, refresh_token) unless refresh_token.nil?
```
chris-rock: I would prefer if we would set the `node['audit']['insecure']` attribute as default to `true` and pass in the required attribute.

libraries/helper.rb
chris-rock: Lets use false as default and pass in `node['audit']['token']` from the calling function

chris-rock:
@jeremymv2 This is a really good improvement. Just some nitpicking comments :-) LGTM once you addressed the issues

libraries/profile.rb
srenatus: @chris-rock: you meant `false`, right? (guessing from the comment above)

jeremymv2: @chris-rock @srenatus I agree completely.  I've added a `node['audit']['insecure']` attribute and set the default to `false` in the library.

chris-rock:
Awesome @jeremymv2 Can you rebase this improvement?

chris-rock:
will be handled with #80

PR #71, opened 2016-06-29
Migrate to chef_handler from resource-based audit

### LINKED ISSUE #70: audit cookbook compliance run and report should not report converge
### Scenario:
We still can't run the `audit` cookbook in production because it reports resources that are converged, which causes us to lose the ability to know when our run is entirely idempotent.

### Steps to Reproduce:
Run the `audit` cookbook configured to run against profiles

### Expected Result:
At the end of an `audit` cookbook run, 0 resources should report as failed

### Actual Result:
resources report as converged

### Implementation 

The `audit` cookbook is implemented incorrectly. Resources in chef are there to configure the machine, not report on that machine. This is a fundamental misuse of the chef model.

Instead a [report handler](https://docs.chef.io/handlers.html) should be used to both download and report on the chef run. The recipe can ensure that inspec is properly installed and that the handler is used.

I am happy to create a PR for this if you think it's the right direction for the cookbook. It is a fundamental departure from what is there now.

mhedgpeth:
### Description

This is a major refactoring of the cookbook and is preliminary for feedback at this point. It addresses the problem with the current version defined as resources and thus reporting "converges" when doing reporting. Instead, this implements a `chef_handler` that does reporting, which fits the pattern much better.

### Issues Resolved

#70 

### Check List
- [X] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD


.kitchen.yml
```diff
@@ -1,78 +1,37 @@
 ---
 driver:
-  name: dokken
+  name: vagrant
```
bigbam505: This seems like a step backwards? Or was this just for your testing?

mhedgpeth: I couldn't run `dokken` on my windows machine. Plus I'm not convinced this ever worked, at least for the chef server. So it was for my testing. I'd be happy to roll it back, but don't see `dokken` as adding any benefit.

bigbam505: Usually docker tests are faster / I think you can run it in travis

mhedgpeth: Yeah so faster tests + CI but no windows development support.I would understand if the former was deemed more valuable.

chris-rock: I propose to add multiple kitchen files as we are doing in https://github.com/chef/inspec 
Reason: we cannot run vagrant in travis. Still like to see the vagrant option within the cookbook. Just move it to .kitchen.vagrant.yml

mhedgpeth: I reverted back to `dokken` per @chris-rock and @bigbam505 suggestions. There is an alternative for vagrant called `.kitchen.vagrant.yml`

vjeffrey:
ok, I pulled down this branch, resolved the conflicts, worked through a few more small code changes that needed to be done, and this *seems* to be working (tested against compliance server and visibility).  That work is on this branch: https://github.com/chef-cookbooks/audit/compare/vj/try-to-merge-handler    I can push those changes up here directly as well, I just pushed to a separate branch in case I, ya know, messed it all up :)  If ppl could pls test that this is working the way we all expect it to...that would be awesome.

jeremymv2:
I propose we delay this until there is a better plan for fully replacing `audit_mode` with built-in inspec functionality in chef-client. There's just so much happening in this PR that I'm not sure it's worth the effort without very careful planning as to how inspec + chef-client work together.

chris-rock:
@mhedgpeth although we have not merged your PR, it was the inspiration to rewrite and clean up a lot in that cookbook. Again we thank you to demonstrate the use of Chef Handler and how that can help the `audit` cookbook. We mention this effort in our README via #130

We are looking forward to get more feedback and PRs!

PR #85, opened 2016-08-23
improve compliance refresh token handling

chris-rock:
### Description

This PR is doing the following:
- extract inspec gem installation
- ensures the refresh token is only exchanged once

### Issues Resolved

Customers reported that `refresh_token` handling was not working as expected and lead to authentication errors for some requests.

### Check List

### DIFF
--- a/libraries/inspec.rb
+++ b/libraries/inspec.rb
@@ -1,0 +1,38 @@
+# encoding: utf-8
+class Audit
+  class Resource
+    class ChefInspec < Chef::Resource
+      resource_name :inspec
+
+      property :version, String, default: 'latest'
+
+      default_action :install
+
+      # installs inspec if required
+      action :install do
+        converge_by 'install/update inspec' do
+          chef_gem 'inspec' do
+            version new_resource.version if new_resource.version != 'latest'
+            compile_time true
+            action :install
+          end
+        end
+
+        converge_by 'verifies the inspec version' do
+          verify_inspec_version version
+        end
+      end
+
+      def verify_inspec_version(inspec_version)
+        require 'inspec'
+        # check we have the right inspec version
+        if Inspec::VERSION != inspec_version && inspec_version !='latest'
+          Chef::Log.warn "Wrong version of inspec (#{Inspec::VERSION}), please "\
+            'remove old versions (/opt/chef/embedded/bin/gem uninstall inspec).'
+        else
+          Chef::Log.warn "Using inspec version: (#{Inspec::VERSION})"
+        end
+      end
+    end
+  end
+end
--- a/libraries/profile.rb
+++ b/libraries/profile.rb
@@ -6,200 +6,176 @@ require 'fileutils'
 
 # `compliance_profile` custom resource to collect and run Chef Compliance
 # profiles
-class ComplianceProfile < Chef::Resource
-  include ComplianceHelpers
-  use_automatic_resource_name
+class Audit
+  class Resource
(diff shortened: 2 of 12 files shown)


libraries/inspec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      property :version, String, default: 'latest'
+
+      default_action :install
+
+      # installs inspec if required
+      action :install do
+        converge_by 'install/update inspec' do
+          chef_gem 'inspec' do
+            version version if version != 'latest'
```
jeremymv2: @chris-rock I get an WARN on line 15 `version` 

Recipe: audit::_inspec
``` 
 * inspec[inspec] action install[2016-08-25T13:28:53+00:00] WARN: /var/chef/cache/cookbooks/audit/libraries/inspec.rb:15:in `block (3 levels) in <class:ChefInspec>': property version is declared in both chef_gem[inspec] and inspec[inspec] action :install. Use new_resource.version instead. At /var/chef/cache/cookbooks/audit/libraries/inspec.rb:15:in `block (3 levels) in <class:ChefInspec>'
[2016-08-25T13:28:53+00:00] WARN: /var/chef/cache/cookbooks/audit/libraries/inspec.rb:15:in `block (3 levels) in <class:ChefInspec>': property version is declared in both chef_gem[inspec] and inspec[inspec] action :install. Use new_resource.version instead. At /var/chef/cache/cookbooks/audit/libraries/inspec.rb:15:in `block (3 levels) in <class:ChefInspec>'
```
This doc seems to support the WARN message above:
https://docs.chef.io/custom_resources.html#new-resource-property

When I test changing line 15 to the following I no longer get the WARNing.
`version new_resource.version if new_resource.version != 'latest'`

chris-rock: 👍

libraries/profile.rb
```diff
@@ -6,200 +6,178 @@
...
+        converge_by 'fetch compliance profile' do
+          return if path # will be fetched from other source during execute phase
+
+          o, p = normalize_owner_profile
+          Chef::Log.info "Fetch compliance profile #{o}/#{p}"
+
+          path = tar_path
+          access_token = token
+
+          puts "ACCESSTOKEN #{access_token}"
```
jeremymv2: Being a little nit-picky; line 58 looks to be a left-over debugging artifact.  As-is it is appended to the end of current line in console/logs during converges and doesn't have any context and is empty when `['audit']['collector'] == chef-server`.  Perhaps, if this is valuable/desirable info to have in log files (which I'm not totally convinced it is), then a line like: `Chef::Log.info "Using Access Token: #{access_token}" if access_token` would be more delightful for the users?

libraries/report.rb
```diff
@@ -1,108 +1,112 @@
...
+      property :environment, String # default: node.environment
+      property :owner, [String, nil]
+
+      default_action :execute
+
+      action :execute do
+        converge_by "report compliance profiles' results" do
+          reports, ownermap = compound_report(profiles)
+
+          puts "ACCESSTOKEN #{token}"
```
jeremymv2: same comment as before about ACCESSTOKEN output.

jeremymv2:
Added a few comments above.  Otherwise this PR looks awesome and will be super helpful.  Tested the PR itself and works like charm!  Very nice @chris-rock!

libraries/report.rb
chris-rock: 👍 Thanks for highlighting this @jeremymv2

PR #91, opened 2016-09-12
Compliance Token resource

jeremymv2:
### Description

This PR introduces `compliance_token` resource.
The goal is to have a self-contained resource for storing the Compliance Server token and exchanging the `refresh_token` for an access token, if needed.

The `compliance_token` resource stashes the token in Chef's `node.run_state` hash. The token can then be passed between resources, and evaluated during the execution phase. run_state is always discarded at the end of the chef-client run.

### Issues Resolved

N/A

### Check List
- [x] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD
- [x] New functionality includes testing.
- [x] New functionality has been documented in the README if applicable
- [x] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD

PR #92, opened 2016-09-12
profile upload

jeremymv2:
### Description

This PR provides action `upload` on resource `compliance_profile`

It requires PR #91 to be merged first.

Only one mode is supported: direct communication with Compliance Server via token||refresh_token.

Usage is based on the audit hash like so:
```
default['audit']['server'] = 'https://compliance-server.test/api'
default['audit']['collector'] = 'chef-compliance'
default['audit']['token'] = nil
default['audit']['refresh_token'] = '20/cgm927J9371DRh8mZ2wp_bXRJxKPW52rtThkwQz5PzhBIcbq6VLB0LsdrfOXv_aBxprU1LTv3aUvHxe2mR-m2A=='
default['audit']['profiles'] = {
  'admin/ssh2' => {
    'source' => '/some/base_ssh.tar.gz'
  },
  'admin/linux2' => {
    'source' => '/some/base_linux.tar.gz'
  }
}
default['audit']['insecure'] = true
```

The goal is to help enable Compliance Profile uploads via automated pipeline workflow. 

### Issues Resolved

N/A

### Check List

### DIFF
--- a/libraries/profile.rb
+++ b/libraries/profile.rb
@@ -140,6 +140,22 @@ class Audit
         end
       end
 
+      action :upload do
+        converge_by 'upload compliance profile' do
+          fail 'unable to determine path to compressed profile archive' if path.nil?
+          o, p = normalize_owner_profile
+          node.run_state['compliance'] ||= {}
+
+          if node.run_state['compliance']['access_token']
+            reqpath ="owners/#{o}/compliance/#{p}/tar"
+            url = construct_url(server, reqpath)
+            Chef::Log.info "Upload compliance profile #{o}/#{p} to: #{url}"
+            res = Compliance::HTTP.post_file(url.to_s, node.run_state['compliance']['access_token'], path, insecure)
+            Chef::Log.info "Got: #{res.code} #{res.message}"
+          end
+        end
+      end
+
       def normalize_owner_profile
         if profile.include?('/')
           profile.split('/').last(2)
--- a/libraries/profile.rb
+++ b/libraries/profile.rb
@@ -18,7 +18,6 @@ class Audit
       # to use a chef-compliance server that is used with chef-server integration
       property :server, [String, URI, nil]
       property :port, Integer
-      property :token, [String, nil]
       property :insecure, [TrueClass, FalseClass], default: false
       property :formatter, ['json', 'json-min'], default: 'json-min'
       property :quiet, [TrueClass, FalseClass], default: true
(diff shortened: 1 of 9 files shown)


jeremymv2:
If PR #91 passes muster, I'll add unit tests and update the README.md with usage info.

chris-rock:
@jeremymv2 super helpful for build cookbooks

libraries/profile.rb
```diff
@@ -141,6 +141,22 @@ class ComplianceProfile < Chef::Resource
...
+        converge_by 'upload compliance profile' do
+          raise 'unable to determine path to compressed profile archive' if path.nil?
+          o, p = normalize_owner_profile
+          node.run_state['compliance'] ||= {}
+
+          if node.run_state['compliance']['access_token']
+            reqpath ="owners/#{o}/compliance/#{p}/tar"
+            url = construct_url(server, reqpath)
+            Chef::Log.info "Upload compliance profile #{o}/#{p} to: #{url}"
+            res = Compliance::HTTP.post_file(url.to_s, node.run_state['compliance']['access_token'], path, insecure)
```
chris-rock: Currently that is not using the upload function for inspec https://github.com/chef/inspec/blob/master/lib/bundles/inspec-compliance/api.rb#L82 Any specific reason?

We need to be aware that the current method is not using the checking mechanisms as available. https://github.com/chef/inspec/blob/master/lib/bundles/inspec-compliance/cli.rb#L94-L167

Do you think we could combine them? I would make to just reuse functions we already have. If that requires refactoring in InSpec, I would go for that.

jeremymv2:
@chris-rock Thanks!  I'll go ahead and add some tests as well as update the docs with the new functionality.  Then this PR will be ready for review.

libraries/profile.rb
jeremymv2: @chris-rock  I agree, using `Compliance::API.upload()` is definitely the way to go.  I'll make that change easily.

The error checking would be a nice to have.  I'm thinking that refactoring the `inspec-compliance` bundle to pull the error checking out of  `ComplianceCli` class would be the longer term solution. The tricky part being to make it non `cli` specific and more API friendly.

What if as a first stab, we duplicate some of the rudimentary error checking from above in the `audit` cookbook resource? Like:

1. path exist
2. profile.check
3. profile exists? (overwrite should probably be the default for a build-cookbook scenario?)

jeremymv2: `Compliance::API.upload()` calls `Compliance::HTTP.post_file()` like so:

`res = Compliance::HTTP.post_file(url, config['token'], archive_path, config['insecure'], !config.supported?(:oidc))`

However, when I pass in a config from `Compliance::Configuration.new()`, `!config.supported?(:oidc))` always resolves to `true` then authentication fails, when `false` the call succeeds.

```
pry(#<Audit::Resource::ComplianceProfile action provider>)> !config.supported?(:oidc)
=> true
[10] pry(#<Audit::Resource::ComplianceProfile action provider>)> !config.supported?(:bar)
=> true
[11] pry(#<Audit::Resource::ComplianceProfile action provider>)>
```

jeremymv2:
Closing this PR in favor of #96

Issue #98, opened 2016-09-21
Update to InSpec 1.0


arlimus:

arlimus:
Apparently someone over at InSpec decided to change the JSON format:

https://github.com/chef/inspec/pull/811

Let's make sure the visibility reporters can actually work with it :grin:

PR #99, opened 2016-09-21
update to work with inspec 1.0 json format

vjeffrey:
fixes https://github.com/chef-cookbooks/audit/issues/98

updates to 1.0 json format

tested with visibility workflow with this change https://delivery.shd.chef.co/e/Chef/#/organizations/Chef_Delivery/projects/delivery/changes/93fe5e4f-107a-4176-aaa6-0290eb9369ba/status/verify

working on those failing tests real quick -ok, fixed.


vjeffrey:
argh, having trouble figuring out what's up with that last test.

srenatus:
@vjeffrey: one clarification question re: failing tests (controls), besides that, LGTM

libraries/collector_classes.rb
```diff
@@ -90,6 +80,7 @@ def count_controls(profiles)
       return count unless profiles.is_a?(Array)
 
       profiles.each do |profile|
+        profile = profile[0]
```
stevendanna: A comment here explaining the format of profiles and profile might be helpful

alexpop:
Had a quick look at it, but I'll have to drop into the shell as well to test it against automate.

I think this change deserves a new pin here:
https://github.com/chef-cookbooks/audit/blob/v0.14.4/attributes/default.rb#L48

and bump the major version of the cookbook as well.

chris-rock:
@alexpop we can merge it on master but not release it until inspec 1.0 is available, in the meantime we can pin to the beta releases

libraries/report.rb
```diff
@@ -61,8 +55,6 @@ class ComplianceReport < Chef::Resource
           else
             Chef::Log.warn "#{collector} is not a supported inspec report collector"
           end
-
-          raise "#{total_failed} audits have failed.  Aborting chef-client run." if total_failed > 0 && node['audit']['fail_if_any_audits_failed']
```
vjeffrey: we no longer have a summary in the new format, which means we don't have failed counts, ya :/

srenatus: Hmm. Since [this is the only use](https://github.com/chef-cookbooks/audit/search?utf8=%E2%9C%93&q=fail_if_any_audits_failed), at least the attribute should go away. That said, can't we aggregate in the cookbook code?

libraries/collector_classes.rb
vjeffrey: yup, that makes sense! i'll go do that right now.

libraries/report.rb
srenatus: this is just a quick hack, but _if this functionality should be retained_, we could go with something like

```ruby
total_failed = reports.map do |name, report|
  report['profiles'].map do |profile|
    profile['controls'].map do |control|
      control['results'].map do |result|
        result['status'] != 'passed' ? 1 : 0
      end
    end
  end
end.flatten.reduce(:+)
```

vjeffrey: thank youuuu @srenatus !!!

vjeffrey:
@alexpop is there a reason you wanted to pin the inspec version? why not just always grab latest, here and in the viz data generator?

vjeffrey:
working on implementing the fix @srenatus got for us on report.rb; when controls have no tests it fails out, b/c there's no control['results'], so just trying to work around that

alexpop:
With commit https://github.com/chef-cookbooks/audit/pull/99/commits/f50d80a594f955b867283f7b3ffca9446eb5e46d, we are sending the same data structure to Chef Automate. I've updated the Visibility merge request and there are no more javascript changes needed.

vjeffrey:
nice! thanks alex; i just finished testing it all; works great :)
one more question: why are we version pinning inspec and not just doing latest?

vjeffrey:
just pushed up the change to keep the 'fail_if_any_audits_fail' functionality

alexpop:
@vjeffrey we've been pinning the inspec version in `attributes/default.rb` for some time now. It allows us to release the `audit` cookbook with a compatible version of `inspec` pinned.

Users can still override the attribute to a different version or `latest` but it might not work.

srenatus:
Looks good to me :+1:

vjeffrey:
the shd/visibility change was merged a minute ago by some nice reviewers :) so, as soon as we release inspec 1.0 i'll update the version in here and we can merge away

vjeffrey:
updated to 1.0.0

alexpop:
Many thanks and hugs!

PR #103, opened 2016-10-06
Use the new method to retrieve access tokens and fix total_failed bug

alexpop:
* This uses the new method to retrieve access tokens from InSpec plugin
* Fixes null pointer exception when calculating total failed controls


alexpop:
Signed-off-by: Alex Pop apop@chef.io

arlimus:
Looks great, thank you @alexpop !

PR #106, opened 2016-10-12
Fix bug when counting total failed controls in json format

alexpop:
`total_failed` only worked for `json-min` format. The current version was not working on Chef Visibility reporting.


attributes/default.rb
```diff
@@ -47,7 +47,7 @@
 default['audit']['fail_if_any_audits_failed'] = false
 
 # inspec gem version to install(e.g. '1.1.0')
-default['audit']['inspec_version'] = '1.1.0'
+default['audit']['inspec_version'] = '1.2.0'
```
chris-rock: 👍

chris-rock:
Thanks for finding and fixing the issue @alexpop!

PR #113, opened 2016-10-17
use chef handler to run inspec tests

vjeffrey:
so many more things to do. just the beginning...working on viz integration parts now.


vjeffrey:
works with visibility with a change from: 
```
# node.default['audit']['profiles'].merge!({
#   'nathen/tmp_compliance_profile' => {
#     'source' => "#{PROFILES_PATH}/tmp_compliance_profile"
#   }
# })
```
to
`node.default['audit']['profiles'].push("#{PROFILES_PATH}/tmp_compliance_profile")`

examples/kitchen/.kitchen.linux.yml
```diff
@@ -1,37 +0,0 @@
----
```
chris-rock: we should not delete the examples. Instead we should migrate to work with the new version

attributes/default.rb
```diff
@@ -46,11 +46,9 @@
 # fail converge after posting report if any audits have failed
 default['audit']['fail_if_any_audits_failed'] = false
 
-# inspec gem version to install(e.g. '1.1.0')
-default['audit']['inspec_version'] = '1.2.0'
```
chris-rock: we are going to need it for now

recipes/default.rb
```diff
@@ -1,91 +1,24 @@
...
-    formatter formatter
-    server server
-    path path unless path.nil?
-    quiet node['audit']['quiet'] unless node['audit']['quiet'].nil?
-    only_if { profile_overdue_to_run?(p, interval_seconds) }
-    action [:fetch, :execute]
-    notifies :touch, "file[#{compliance_cache_directory}/#{p}]", :delayed
-    notifies :execute, "compliance_report[#{report_collector}]", :delayed
-  end
+chef_inspec 'inspec' do
```
chris-rock: `chef_inspec` feels like a strange name

resources/inspec.rb
```diff
@@ -1,7 +1,7 @@
 # encoding: utf-8
 
 provides :chef_inspec
```
chris-rock: can we just inspec for provides and as a resource name instead of `chef_inspec`?

recipes/default.rb
```diff
@@ -1,91 +1,24 @@
...
 
-# report the results
-compliance_report report_collector do
-  owner node['audit']['owner']
-  server server
-  collector report_collector
-  quiet node['audit']['quiet'] unless node['audit']['quiet'].nil?
-  action :nothing
-end if node['audit']['profiles'].values.any?
+inspec_report 'report' do
```
chris-rock: Do you think we should combine:
```
chef_handler 'Chef::Handler::AuditReport' do
  source "#{handler_directory}/audit_report.rb"
  action :enable
end

inspec_report 'report' do
  action :execute
end
```

vjeffrey: how do we combine them?

recipes/default.rb
```diff
@@ -1,91 +1,26 @@
...
-    formatter formatter
-    server server
-    path path unless path.nil?
-    quiet node['audit']['quiet'] unless node['audit']['quiet'].nil?
-    only_if { profile_overdue_to_run?(p, interval_seconds) }
-    action [:fetch, :execute]
-    notifies :touch, "file[#{compliance_cache_directory}/#{p}]", :delayed
-    notifies :execute, "compliance_report[#{report_collector}]", :delayed
-  end
+chef_handler 'Chef::Handler::AuditReport' do
```
chris-rock: As discussed with @vjeffrey we try to find a way that does not require us to write a file on the host node

.kitchen.yml
```diff
@@ -64,8 +64,6 @@ suites:
     attributes:
       audit:
         profiles: &profiles
```
vjeffrey: leftover from initial implementation.. Yaml syntax docs: "Repeated nodes are first identified by an anchor (marked with the ampersand - “&”), and are then aliased (referenced with an asterisk - “*”) thereafter."

attributes/default.rb
```diff
@@ -46,11 +48,9 @@
 # fail converge after posting report if any audits have failed
 default['audit']['fail_if_any_audits_failed'] = false
```
vjeffrey: sounds good to me! i'll remove it

vjeffrey:
still need to figure out the chef_handler putting a source file on the system thing (https://github.com/chef-cookbooks/audit/pull/113/files#diff-55bf87238c9b8af164c5133a60721f12R9)  
I've asked around...ppl are unsure of what to do about that. one suggestion: "maybe the `chef_handler` resource can delete the file if you use the `:disable` action. "

jeremymv2:
Most handlers that I've used lay down the handler libs as a versioned gem.  One of the better examples is the DataDog handler: https://github.com/DataDog/chef-datadog/blob/master/recipes/dd-handler.rb#L32-L38

vjeffrey:
I've been playing around with trying to just do a cleanup recipe that disables the chef_handler (in hopes that it would clean things up), or, alternatively, straight-up deletes the handler directory.  unfortunately, i haven't had any luck. 

we're doing what's described in the docs atm (http://chef.readthedocs.io/en/latest/essentials_handlers.html, default.rb)

vjeffrey:
yup. so we were trying to avoid the gem thing in the case of an air-gapped env/to avoid outside deps (I know that doesn't make much sense right now since we are installing inspec gem but we are trying to find a way to move away from that and vendor) --- I spoke with joshua timberman and jj ashgar and both said ya, those are the only two ways of doing the chef handler thing, so i'm just gonna whip up a quick delete directory cleanup recipe and include it at the end of default with a good ol include_recipe call.  coming soon!  just have to make food for ppl real quick, then i'll do it :)

recipes/clean-up.rb
```diff
@@ -0,0 +1,7 @@
+
+handler_directory = ::File.join(Chef::Config[:file_cache_path], 'handler')
+
+directory handler_directory do
+  recursive true
+  action :delete
+end
```
jeremymv2: I see what you're doing now :)  However, I suppose I don't understand what the requirement is to cause us to remove the chef handler ruby library code.
To me it smells fishy because it seems like it's akin to running an app that deletes all traces of itself after execution.  I can think of two potential concerns that arise from that:
- It's not in the spirit of transparency - we're running code on nodes and then deleting that code from the nodes.
- Deleting the handler code from the node could make troubleshooting harder since it requires version correlation against a git repo or chef server to determine what actual code was running when the handler fired.  Having the code on intact on the node gives you a 100% assurance that you know what executed.    

If this moves forward, I recommend adding the standard header to the recipe which gets generated from `chef generate recipe clean-up`  One last thing on this commit, just me perhaps, but using `-` in recipe names is not consistent with the requirement that cookbooks are not allowed to have `-` in the name.

```
#
# Cookbook Name:: audit
# Recipe:: clean-up
#
# Copyright (c) 2016 The Authors, All Rights Reserved.
```

chris-rock: @jeremymv2 thanks for the honest feedback. That is very helpful! Background: This was an approach @vjeffrey and I discussed before. From my perspective it feels strange to install code on a node just for runtime and not cleaning the tmp stuff up. Nevertheless if that is not standard practice we should not do this approach.

@vjeffrey I propose we leave the handler files there for now, I've an idea how we can get around generating the file in future to get the best of both worlds. But lets try that in future PRs.

vjeffrey: oooh good points! i didn't know about the naming thing, so i'll definitely clean that up if we go forward with this.  i'm not certain we need to be deleting anything, I was just trying to make things cleaner....those are some good issues you brought up... @chris-rock, what's your opinion on this?

vjeffrey: cool, sounds good! also, get some sleep @chris-rock!!

vjeffrey: just updated by dropping that last commit.

PR #114, opened 2016-10-19
Activate test-kitchen in travis

chris-rock:
this PR adds chef-client runs via travis


alexpop:
Looks good Chris. Are we adding a profile to execute and some kitchen verify tests later on?

chris-rock:
@alexpop yes, we will do this, once we implemented https://github.com/chef/inspec/issues/1227 for audit cookbook

PR #120, opened 2016-10-25
Update Github PR template

tas50:
Update the template for the DCO


tas50:
Obvious fix that doesn't actually require the DCO

Issue #118, opened 2016-10-24
Implement RFC: Harmonize profile location targets


chris-rock:
### Cookbook version

2.x

### Scenario:

It should be easy for customer to copy profile locations from `inspec.yml` to chef attributes. In https://github.com/chef/inspec/issues/1227 we defined the target format for `audit` cookbook. We change from

```
"profiles" => {
    # org / profile name from Chef Compliance
    "base/linux" => true,
    # supermarket url
    "brewinc/ssh-hardening" => {
      # location where inspec will fetch the profile from
      "source" => "supermarket://hardening/ssh-hardening"
    },
    # local Windows path
    "brewinc/win2012_audit" => {
      # filesystem path
      "source" => "E:/profiles/win2012_audit"
    },
    # github url
    "brewinc/tmp_compliance_profile" => {
      "source" => "https://github.com/nathenharvey/tmp_compliance_profile"
    },
    # disable profile
    "brewinc/tmp_compliance_profile-master" => {
      "source" => "/tmp/tmp_compliance_profile-master",
      "disabled" => true
    }
  }
```

to

```
{
  "profiles": [
    {
      "name": "hardening/ssh-hardening"
    },
    {
      "name": "os-hardening",
      "url": "https://github.com/dev-sec/tests-os-hardening/archive/master.zip"
    },
    {
      "git": "https://github.com/dev-sec/tests-ssh-hardening.git"
    },
    {
      "git": "https://github.com/dev-sec/tests-os-hardening.git"
    }
  ]
}
```

vjeffrey:
```profiles = {
  "profiles": [
    {
      "name": "hardening/ssh-hardening"
    },
    {
      "name": "os-hardening",
      "url": "https://github.com/dev-sec/tests-os-hardening/archive/master.zip"
    },
    {
      "git": "https://github.com/dev-sec/tests-ssh-hardening.git"
    },
    {
      "git": "https://github.com/dev-sec/tests-os-hardening.git"
    }
  ]
}

profiles_array = []

profiles[:profiles].each { |profile| profiles_array.push(profile[:url] || profile[:git]) }

profiles[:profiles].each { |profile|
  if profile.length == 1 && profile[:name]
    profiles_array.push("supermarket://" + profile[:name])
  end
}

profiles[:profiles].each { |profile|
  if profile.include?(:supermarket_url)
    profiles_array.push(profile[:supermarket_url] + "/" + profile[:name])
  end
}

[...]

PR #121, opened 2016-10-25
 harmonize profile targets

vjeffrey:
fixes: https://github.com/chef-cookbooks/audit/issues/118

unsure about the disabled part, looking into that to confirm that works..

### DIFF
--- a/Berksfile
+++ b/Berksfile
@@ -2,3 +2,6 @@
 source 'https://supermarket.chef.io'
 
 metadata
+
+cookbook 'apt'
+cookbook 'yum'
--- a/README.md
+++ b/README.md
@@ -110,33 +110,41 @@ This requires your Chef Server to be integrated with the Chef Compliance server 
 
 Once the cookbook is available in Chef Server, you need to add the `audit::default` recipe to the run-list of each node. The profiles are selected via the `node['audit']['profiles']` attribute. For example you can define the attributes in a role or environment file like this:
 
+
+node.default['audit']['profiles'].push("source": "#{PROFILES_PATH}/mylinux-failure-success")
+
 ```ruby
 "audit" => {
   "collector" => "chef-server",
   "inspec_version" => "1.2.1",
-  "profiles" => {
-    # org / profile name from Chef Compliance
-    "base/linux" => true,
-    # supermarket url
-    "brewinc/ssh-hardening" => {
-      # location where inspec will fetch the profile from
-      "source" => "supermarket://hardening/ssh-hardening"
+  "profiles" => [
+    # profile from Chef Compliance
+    {
+      "name": "linux",
+      "compliance": "base/linux"
+    },
+    # profile from supermarket
+    {
+      "name": "ssh",
+      "supermarket": "hardening/ssh-hardening"
     },
     # local Windows path
-    "brewinc/win2012_audit" => {
+    {
+      "name": "brewinc/win2012_audit",
       # filesystem path
-      "source" => "E:/profiles/win2012_audit"
+      "source": "E:/profiles/win2012_audit"
     },
-    # github url
-    "brewinc/tmp_compliance_profile" => {
-      "source" => "https://github.com/nathenharvey/tmp_compliance_profile"
+    # github
+    {
+      "name": "ssl",
+      "git": "https://github.com/dev-sec/ssl-benchmark.git"
     },
-    # disable profile
-    "brewinc/tmp_compliance_profile-master" => {
-      "source" => "/tmp/tmp_compliance_profile-master",
-      "disabled" => true
+    # url
+    {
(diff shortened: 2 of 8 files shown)


chris-rock:
Lets remove the disabled part, not sure why we introduced that in the first place. You can just remove the block for the profile location if you do not want to execute it. If I remember correctly that was intended to load profiles but do not report them. We do not need that anymore, since we will rely on InSpec core to handle that.

libraries/helper.rb
```diff
@@ -23,4 +23,31 @@ def entity_uuid
...
   end
+
+  def get_tests_for_runner
+    profiles = node['audit']['profiles']
+    profiles_array = []
+    profiles.each { |profile| profiles_array.push(profile[:url] || profile[:git] || profile[:source]) }
+
+    profiles.each { |profile|
+      if profile.length == 1 && profile[:name]
+        profiles_array.push("supermarket://" + profile[:name])
```
chris-rock: I think we want to give the runner the whole hash directly instead of transforming that to urls

chris-rock: we would need to adapt the `add_target` method in inspec: https://github.com/chef/inspec/blob/master/lib/inspec/runner.rb#L156-L164 to receive objects. By default it is using that function to change from url string to hash object: https://github.com/chef/inspec/blob/master/lib/inspec/profile.rb#L43-L46.

Instead we would invoke https://github.com/chef/inspec/blob/master/lib/inspec/profile.rb#L38-L41 from the runner initially. See how it is invoked at https://github.com/chef/inspec/blob/master/lib/inspec/dependencies/requirement.rb#L101-L106

vjeffrey:
ok, so with this change to the runner we can accept everything except the compliance and supermarket profiles: 
```
    def add_target(target, _opts = [])
      if target.is_a?(Array)
        target.each { |profile_target| add_single_target(profile_target) }
      else
        add_single_target(target)
      end
    end

    def add_single_target(target)
      profile = Inspec::Profile.for_target(target,
                                           cache: @cache,
                                           backend: @backend,
                                           controls: @controls,
                                           attributes: @conf[:attributes])
      fail "Could not resolve #{target} to valid input." if profile.nil?
      @target_profiles << profile if supports_profile?(profile)
    end
```
screen shot 2016-10-25 at 9 57 04 am

vjeffrey:
working on the compliance and supermarket ones now..... :/

vjeffrey:
tried to get test cookbook to include in examples folder for this, but having trouble.... ` Waiting for SSH service on 127.0.0.1:2200, retrying in 3 seconds` i'll work on getting that in the morning...not sure what's happening there

README.md
```diff
@@ -110,33 +110,41 @@ This requires your Chef Server to be integrated with the Chef Compliance server
...
+      "name": "ssh",
+      "supermarket": "hardening/ssh-hardening"
     },
     # local Windows path
-    "brewinc/win2012_audit" => {
+    {
+      "name": "brewinc/win2012_audit",
       # filesystem path
-      "source" => "E:/profiles/win2012_audit"
+      "source": "E:/profiles/win2012_audit"
```
chris-rock: i think it needs to be `path` now

vjeffrey: oh shoot, yes it does need to be path!

vjeffrey:
@chris-rock: it's ready!

chris-rock:
Thanks @vjeffrey for this great improvement

Berksfile
```diff
@@ -2,3 +2,6 @@
 source 'https://supermarket.chef.io'
 
 metadata
+
+cookbook 'apt'
+cookbook 'yum'
```
jeremymv2: It looks like 'git' is just going to be used during testing/validation.  In that case, the better pattern is to use a fixture cookbook to install git rather than a 'prep' recipe.  See this for example of install 'java' for testing the 'jenkins' cookbook: 
https://github.com/chef-cookbooks/jenkins/blob/master/Berksfile#L9-L21

https://github.com/chef-cookbooks/jenkins/blob/master/test/fixtures/cookbooks/jenkins_server_wrapper/recipes/default.rb#L3

recipes/prep.rb
```diff
@@ -0,0 +1,2 @@
+# this is required to run git dependencies via inspec
+package 'git'
```
jeremymv2: This recipe seems out of place if only for the purpose of being used during the testing phase in kitchen of this cookbook.  In my opinion, a better approach is to install it via a fixture cookbook.

Berksfile
jeremymv2: A simple/safe way of installing the git binary on linux is via the git cookbook:  

https://github.com/chef-cookbooks/git/blob/master/recipes/default.rb

PR #122, opened 2016-10-26
Upload profiles to Chef Compliance via Chef resource

vjeffrey:
tada!  it works  @chris-rock  .....i have a bunch of requires in the inspec resource....not sure how to get around that?

using this to test it: https://github.com/vjeffrey/testing-audit

### DIFF
--- a/recipes/upload.rb
+++ b/recipes/upload.rb
@@ -27,10 +27,10 @@ end
 node['audit']['profiles'].each do |profile|
   profile_owner = profile[:name]
   profile_path = profile[:path]
-  raise "Invalid profile name '#{profile_owner}'. "\
+  Chef::Log.error "Invalid profile name '#{profile_owner}'. "\
        "Must contain /, e.g. 'john/ssh'" if profile_owner !~ %r{\/}
   _o, p = profile_owner.split('/').last(2)
-  raise "Invalid path '#{profile_path}'" if profile_path.nil?
+  Chef::Log.error "Invalid path '#{profile_path}'" if profile_path.nil?
 
   # upload profile
   compliance_upload p do
--- a/recipes/upload.rb
+++ b/recipes/upload.rb
@@ -1,0 +1,49 @@
+# encoding: utf-8
+#
+# Cookbook Name:: audit
+# Recipe:: upload
+#
+# Copyright 2016 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


resources/inspec.rb
```diff
@@ -22,6 +29,49 @@
...
+  end
+
+  converge_by 'upload compliance profile' do
+    require 'bundles/inspec-compliance/configuration'
+    require 'bundles/inspec-compliance/support'
+    require 'bundles/inspec-compliance/http'
+    require 'bundles/inspec-compliance/api'
+    require 'bundles/inspec-compliance/target'
+    _success, msg, access_token = Compliance::API.get_token_via_refresh_token(server, refresh_token, true)
+
```
jeremymv2: This was solved in the past since there was a compliance library.  All libraries get compiled and automatically included in the cookbook resources.  That would perhaps be a decent place for all the requires and the compliance API methods.

vjeffrey: :+1: sounds like a plan!

recipes/upload.rb
```diff
@@ -0,0 +1,49 @@
...
+  profile_owner = profile[:name]
+  profile_path = profile[:path]
+  raise "Invalid profile name '#{profile_owner}'. "\
+       "Must contain /, e.g. 'john/ssh'" if profile_owner !~ %r{\/}
+  _o, p = profile_owner.split('/').last(2)
+  raise "Invalid path '#{profile_path}'" if profile_path.nil?
+
+  # upload profile
+  inspec p do
+    profile_name p
```
jeremymv2:
```
inspec p do
   profile_name p
```

This feels a little strange to me since the resource is `inspec` - to me that would on the surface mean that it controls the installation and configuration of `inspec` (gem/tools).  This might just be pernickety, but mixing in inspec profiles uploads might make more sense in a separate resource perhaps.

vjeffrey: ya, that makes sense. :+1:

chris-rock:
@vjeffrey Do you intend to add the upload example to our example directory?

resources/upload.rb
```diff
@@ -0,0 +1,54 @@
...
+
+# upload profile to compliance server
+action :upload do
+  converge_by 'run profile validation checks' do
+    load_inspec_libs
+    raise 'Path to profile archive not specified' if path.nil?
+    raise "Profile archive file #{path} does not exist." unless ::File.exist?(path)
+    profile = Inspec::Profile.for_target(path, {})
+    error_count = 0
+    lambda { |msg|
```
chris-rock: Is this lamda used somewhere?

chris-rock: Would be great to tie this warnings and errors to Chef::Log.error and Chef::Log.warn

vjeffrey: i think that lambda was just cruft leftover. i'll remove and test in a bit.   oh ya, good point on the errors.  i'll go through and update those! :)

examples/profile_upload/chefignore
```diff
@@ -0,0 +1,107 @@
+# Put files/directories that should be ignored in this file when uploading
```
chris-rock: lets remove this file

vjeffrey:
@chris-rock: updated!

chris-rock:
Awesome @vjeffrey Lets cover the unit tests in another PR #128

Issue #128, opened 2016-10-27
Add unit tests


chris-rock:
https://github.com/chef-cookbooks/audit/pull/122 re-added support for compliance profile upload. This PR has not covered the unit tests and we should add enable them for 2.0 release.

jeremymv2:
This is what the previous upload unit tests looked like: https://github.com/jeremymv2/audit/blob/7c1930a46dcda3aa0b16e0ae2749d8b3f9a5edd5/spec/unit/recipes/upload_spec.rb

chris-rock:
This should clean up all unit tests:
- for profile upload

We should have a separate file for component and visibility should move from helpers as well.

PR #131, opened 2016-10-28
Add unit tests

vjeffrey:
This commit adds more testing!

A few of them I couldn't get to work correctly....

### DIFF
--- a/files/default/audit_report.rb
+++ b/files/default/audit_report.rb
@@ -15,6 +15,7 @@ class Chef
         interval_enabled = node['audit']['interval']['enabled']
         interval_time = node['audit']['interval']['time']
         report_file = node['audit']['output']
+        profiles = node['audit']['profiles']
 
         load_needed_dependencies
 
@@ -22,8 +23,8 @@ class Chef
         login_to_compliance(server, user, token, refresh_token) if reporter == 'chef-compliance'
 
         if check_interval_settings(interval, interval_enabled, interval_time, report_file)
-          call(reporter)
-          send_report(reporter, server, user)
+          call(reporter, profiles)
+          send_report(reporter, server, user, profiles)
         else
           Chef::Log.error 'Please take a look at your interval settings'
         end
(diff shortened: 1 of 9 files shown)


files/default/audit_report.rb
```diff
@@ -15,15 +15,16 @@ def report
         interval_enabled = node['audit']['interval']['enabled']
         interval_time = node['audit']['interval']['time']
         report_file = node['audit']['output']
+        test_array = node['audit']['profiles']
```
chris-rock: Is `profiles` maybe a better name? I was thinking about an array of single tests first

spec/unit/libraries/default_spec.rb
```diff
@@ -20,6 +20,9 @@
 require 'spec_helper'
```
chris-rock: We may should rename the spec to `visibility_spec.rb`. Do we intent to add more specs for other reporters?

files/default/audit_report.rb
vjeffrey: yup! :+1:

spec/unit/libraries/compliance_spec.rb
```diff
@@ -0,0 +1,26 @@
+# encoding: utf-8
+#
+# Cookbook Name:: audit
+# Spec:: compliance
+
+require 'spec_helper'
+require_relative '../../../libraries/collector_classes'
+
+# TODO: can't get this to work: `# ./libraries/collector_classes.rb:251:in `node_info'`
+# need to figure out how to mock the run context
```
vjeffrey: :( can't figure out how to mock the run context..is there another way around this?

chris-rock: The solution is the removal of the run context. I'll work on a PR to cover that

vjeffrey: thank you!!

chris-rock:
This fixes https://github.com/chef-cookbooks/audit/issues/128

PR #133, opened 2016-10-28
do not hand over run context into reporter

chris-rock:


files/default/audit_report.rb
```diff
@@ -94,12 +94,26 @@ def call(reporter)
...
+        }
+      end
+
       def send_report(reporter, server, user)
         Chef::Log.info "Reporting to #{reporter}"
 
         # TODO: harmonize reporter interface
         if reporter == 'chef-visibility'
-          Collector::ChefVisibility.new(entity_uuid, run_id, run_context.node.name).send_report
+          Collector::ChefVisibility.new(entity_uuid, run_id, gather_nodeinfo).send_report
```
vjeffrey: i think only node name was being sent to viz, which is why the tests are complaining

vjeffrey:
🌟 🎉 :+1:  yay!!!!!   viz tests are breaking cause it only expects the node name, but +:1: once that's fixed! :)

Issue #126, opened 2016-10-27
JSON file reporter


chris-rock:
We should enable the audit cookbook to write json reports into a specific directory. By default the nameing of a report should use the pattern `{profilename}-{timestamp}.json`

vjeffrey:
:+1: this is also needed to have a way to track the interval reporting. i might do 132 + 126 at the same time to avoid breaking intervals in the inbetween

Issue #132, opened 2016-10-28
Remove temporary report file


chris-rock:
### Cookbook version

2.0.0
### Scenario:

Previously we used the runner with generating a json file: https://github.com/chef-cookbooks/audit/blob/v1.1.0/resources/profile.rb#L121-L135 (edited)

```
output = quiet ? ::File::NULL : $stdout
runner = ::Inspec::Runner.new('report' => true, 'format' => formatter, 'output' => output)
report = runner.report.to_json
```

This helps us to remove https://github.com/chef-cookbooks/audit/blob/master/files/default/audit_report.rb#L86-L87 and https://github.com/chef-cookbooks/audit/blob/master/libraries/helper.rb#L74-L80. For users who require a local json report, we should implement https://github.com/chef-cookbooks/audit/issues/126

PR #134, opened 2016-10-29
fix reporting files

vjeffrey:
This commit changes the default behavior to save the json report to a variable that we send over to the collector.
It also enables writing the json report to a file if the interval settings are enabled or the user sets the new default attribute write to file to true

In #126 it was specified that the filename we write to should be created with name and timestamp... this is being done when the write_to_file attribute is set to true, but not when the interval is enabled (for filename-matching purposes)

fixes https://github.com/chef-cookbooks/audit/issues/126 & https://github.com/chef-cookbooks/audit/issues/132

also does some cleanup


README.md
```diff
@@ -343,6 +343,31 @@ You can enable the interval and set the interval time, along with your desired p
 
 ```
 
+## Write to file
+
+If you would like to write the json report to a file on disk, you can enable the write to file attribute.
+  Note:  If write to file is enabled, interval timing may not be enabled.  This is because when we write
```
chris-rock: Could we just touch a file to change its timestamp? Interval handling should not depend on json reporter

vjeffrey:
i know we said we werent' doing multiple reporters yet, but while i was refactoring i thought i could try an idea really quick, and it worked....if we just set the collector as an array, we can stick a line in the audit report after load needed dependencies/before login to compliance that iterates through the collectors....i'm gonna push up a commit with that work in a bit, just in case we decide we want to enable that functionality. can always drop the commit if we dont :)

chris-rock:
@vjeffrey Sounds great. It should not hurt to support that

libraries/helper.rb
```diff
@@ -69,19 +69,50 @@ def base_chef_server_url
...
+    timestamp_file.puts(timestamp)
+    timestamp_file.close
+  end
+
+  # takes value of reporters and returns array to ensure backwards-compatibility
+  def handle_reporters(reporters)
+    return reporters if reporters.is_a? Array
+    new_array = []
+    new_array.push(reporters)
+    new_array
```
chris-rock: maybe just `[reporters]`

files/default/audit_report.rb
```diff
@@ -6,27 +6,44 @@ class Handler
...
+        # create a file with a timestamp to
+        create_timestamp_file if interval_enabled
 
+        # load inspec, supermarket bundle and compliance bundle
         load_needed_dependencies
 
-        # ensure authentication for Chef Compliance is in place
-        login_to_compliance(server, user, token, refresh_token) if reporter == 'chef-compliance'
+        # iterate through reporters
+        reporters.each do |reporter|
```
chris-rock: it would be perfect to iterate only over `send_report(reporter, server, user, profiles)`, but we need `json-min` for Chef Compliance ... so it looks like we need to iterate over the scan as well for now

libraries/collector_classes.rb
```diff
@@ -271,4 +274,34 @@ def send_report
...
+      @timestamp = timestamp
+    end
+
+    def send_report
+      Chef::Log.warn 'Writing report to file.'
+      write_to_file(@report, @profiles, @timestamp)
+    end
+
+    def write_to_file(report, profiles, timestamp)
+      filename = extract_profile_names(profiles) << '-' << timestamp << '.json'
```
chris-rock: I would just use `inspec-timestamp.json`.

libraries/helper.rb
```diff
@@ -69,19 +69,50 @@ def base_chef_server_url
...
 
-  # get file contents where inspec results were saved
-  def results
-    result_path = File.expand_path('../../inspec_results.json', __FILE__)
-    file = File.open(result_path, 'rb')
-    content = file.read
-    file.close
-    content
+  # returns string of profile names separated with underscore
+  def extract_profile_names(profiles)
```
chris-rock: I think we do not need this function...

libraries/collector_classes.rb
vjeffrey: :+1:

libraries/helper.rb
vjeffrey: oh, ha! silly me :)

libraries/helper.rb
vjeffrey: :+1:

vjeffrey:
@chris-rock updated

chris-rock:
nice @vjeffrey

files/default/audit_report.rb
```diff
@@ -44,55 +61,27 @@ def load_needed_dependencies
...
+        Chef::Log.debug "Options are set to: #{opts}"
         runner = ::Inspec::Runner.new(opts)
 
+        # parse profile hashes for runner, see libraries/helper.rb
         tests = tests_for_runner(profiles)
         tests.each { |target| runner.add_target(target, opts) }
 
         Chef::Log.info "Running tests from: #{tests.inspect}"
         runner.run
+        runner.report.to_json
```
chris-rock: +1

chris-rock:
Awesome work @vjeffrey

Issue #129, opened 2016-10-27
Add Chef Server authentication support


chris-rock:
### Cookbook version

2.0.0
### Implementation

In order to address those issues and harmonize the implementation between audit cookbook and inspec, we need to write InSpec fetcher for Chef Server (to have Chef Server authentication), this implicitly removes the need to download profiles manually before execution and also removes the need to aggregate the reports, since InSpec is already doing that out-of-the-box

PR #135, opened 2016-11-01
implement chef-server fetcher and reporter

chris-rock:
This PR implements #129

### DIFF
--- a/files/default/audit_report.rb
+++ b/files/default/audit_report.rb
@@ -26,6 +26,9 @@ class Chef
         # load inspec, supermarket bundle and compliance bundle
         load_needed_dependencies
 
+        # detect if we run in a chef client with chef server
+        load_chef_fetcher if reporters.include?('chef-server')
+
         # iterate through reporters
         reporters.each do |reporter|
           # ensure authentication for Chef Compliance is in place, see libraries/compliance.rb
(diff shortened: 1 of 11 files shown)


files/default/audit_report.rb
```diff
@@ -98,6 +116,22 @@ def gather_nodeinfo
         }
       end
 
+      # this is a helper methods to extract the profiles we scan and hand this
+      # over to the reporter in addition to the `json-min` report. `json-min`
+      # reports do not include information about the source of the profiles
+      # TODO: should be available in inspec `json-min` reports out-of-the-box
+      # TODO: raise warning when not a compliance-known profile
+      def cc_profile_index(profiles)
+        cc_profiles = tests_for_runner(profiles).map { |profile| profile[:compliance] if profile[:compliance] }.uniq.compact
```
srenatus: I think this could be `tests_for_runner(profiles).select { |profile| profile[:compliance] }.uniq.compact`:

```
irb(main):001:0> ls = [ { foo: nil }, { bar: 3 }, { foo: 4 }]
=> [{:foo=>nil}, {:bar=>3}, {:foo=>4}]
irb(main):002:0> ls.select {|x| x[:foo] }
=> [{:foo=>4}]
```

chris-rock: In this case we would need:

`tests_for_runner(profiles).select { |profile| profile[:compliance] }.map { |profile| profile[:compliance].uniq.compact`

I think it reads better, although we loop more often. Thanks @srenatus

metadata.rb
```diff
@@ -5,7 +5,7 @@
 license 'Apache 2.0'
 description 'Allows for fetching and executing compliance profiles, and reporting its results'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version '2.0.0'
+version '2.0.2'
```
chris-rock: do not change the version in this PR

vjeffrey:
:+1: on the code, testing against viz -- have to reboot real quick cause everything is hanging

files/default/audit_report.rb
```diff
@@ -98,6 +115,22 @@ def gather_nodeinfo
...
+      def cc_profile_index(profiles)
+        cc_profiles = tests_for_runner(profiles).select { |profile| profile[:compliance] }.map { |profile| profile[:compliance] }.uniq.compact
+        cc_profiles.map { |profile|
+          owner, profile_id = profile.split('/')
+          {
+            owner: owner,
+            profile_id: profile_id,
+          }
+        }
+      end
```
vjeffrey: would prob be cool to have a test on this method someday... :)

vjeffrey:
tested against viz -- all good :+1:

files/default/audit_report.rb
vjeffrey: also, it could be moved to helper.rb

PR #145, opened 2016-11-04
enable chef-server fetcher attribute

chris-rock:
This allows users to enable the chef-server fetcher with another reporter, e.g. Visibility

Just set the attribute `default['audit']['fetcher'] = 'chef-server'`

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -22,6 +22,12 @@ default['audit']['inspec_version'] = '1.2.0'
 # chef-visibility requires inspec version 0.27.1 or above
 default['audit']['collector'] = 'chef-server'
 
+# It will use an InSpec fetcher that fetches compliance profiles via Chef Server
+# from Chef Compliance. Will be activated by default if the 'chef-server' collector
+# is used
+# fetcher possible values: chef-server
+default['audit']['fetcher'] = nil
+
 # Attributes server, insecure and token/refresh_token are only needed for the 'chef-compliance' collector
 # server format example: 'https://comp-server.example.com/api'
 default['audit']['server'] = nil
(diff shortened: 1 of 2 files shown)


vjeffrey:
🌟  so much awesomenesssssss....i haven't tested, but it's pretty straightforward.  i'll test in a bit if no one gets to it before me

sc0ttruss:
Tested this locally with 9 nodes, used the following role to test whilst, setting the following two lines in the nodes client.rb.   Able to see data in both compliance and in visibility.  Timestamps in both confirm they come from the same origin.   
Output: Major, 58, Minor, 14, Compliant 27, skipped 0, critical 8, - Chef Automate

Whilst it works in automate, and the data seems complete.  In Compliance, the timestamp is correct on the "dashboard" screen ( 17:37 in this case ), but once clicked on the node (scan) in the dashboard, the data is from the previous scan ( 17:11 ).  So, something amiss on the compliance front, unless I have misconfigured something there.   Will check further but update for now. 

```
{
  "name": "audit",
  "chef_tupe": "role",
  "json_class": "Chef::Role",
  "default_attributes": {
    "audit": {
      "profiles": [
      {
        "name": "ssh",
        "compliance": "base/ssh"
      },
      {
        "name": "linux",
        "compliance": "base/linux"
      }
    ],
     "collector": ["chef-visibility","chef-server"]
    }
  }
}
```

and client.rb has the two lines added :note change the hostname if you are copying and the token is the default token for automate.

```
data_collector.server_url "https://automate.myorg.chefdemo.net/data-collector/v0/"
data_collector.token "93a49a4f2482c64126f7b6015e6b0f30284287ee4054ff8807fb63d9cbd1c506"
```

chris-rock:
Thanks @vjeffrey and @sc0ttruss for testing

PR #152, opened 2016-11-08
add reference to self-signed certs with visibility

chris-rock:
Thanks @jerryaldrichiii for highlighting that issue.


vjeffrey:
:+1:

Issue #158, opened 2016-11-16
Improve cookbook usability(fetcher, reporter) renaming


alexpop:
As a follow up to: https://github.com/chef-cookbooks/audit/issues/156

Let's improve cookbook usability by:
* introducing a `chef-server-compliance` collector/reporter and a `chef-server-compliance` fetcher. Log warnings to be displayed when the legacy `chef-server` collector and fetchers are still used.
* the `collector` attribute should also be renamed to `reporter` and log warning messages should be displayed when `collector` is used

PR #163, opened 2016-11-17
Update fetcher for chef-server-visibility and add chef-server-compliance collector

alexpop:
Fixes https://github.com/chef-cookbooks/audit/issues/158

The chef-server fetcher is also using a different fetching URL when reporter chef-server-visibility is used.

I was also thinking to introduce replace the current `chef-server` fetcher with `chef-server-compliance` and possibly add a `chef-server-automate` reporter. In the interest of keeping this change small for now. I reverted the fetcher change.

Attributes and README are not fully updated. Waiting for 👍  before doing that.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,3 +69,5 @@ Style/SpaceAroundOperators:
   Enabled: false
 Style/IfUnlessModifier:
   Enabled: false
+Style/AccessorMethodName:
+  Enabled: false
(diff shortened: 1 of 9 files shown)


vjeffrey:
can we add the automate version requirements to the readme in this pr too, please? so we're clear about what can be used with what versions?

alexpop:
Thanks Stephan, updated based on your feedback!
In the interest of getting the `chef-server-visibility` collector to work, merging this and will sync up with @chris-rock later regarding the `collector` deprecation.

Issue #164, opened 2016-11-17
ability to install inspec as a package


vjeffrey:
As noted by @stephenlauck, there are times (at enterprise or other) when accessing a gem is not an option. We should support installation via package by means of an attribute that decides whether we need to run the inspec recipe or an alternative recipe, e.g. inspec_package

chris-rock:
InSpec is part of Chef Client 13, therefore an additional package is not required anymore.

Issue #112, opened 2016-10-14
Vendor InSpec gem


chris-rock:
Instead of requiring Chef to install the InSpec gem, we vendor the gem into the cookbook. This allows users to update easier.
### Cookbook version

1.0.0

jeremymv2:
@chris-rock 
The vendoring approach can be problematic for a couple of reasons that come to mind:
1. cookbook size - as of inspec 1.3.0 the size is 27MB of core libs + all deps
2. typically you'd manipulate ruby's $LOAD_PATH to prepend your custom vendor lib path from the cookbook; however, if certain gem libs are already present from Chef-client's lib path, the gem library from the chef-client install is ALWAYS used instead of the vendored one. This is because the chef-client already has the gem activated prior to loading/activating the one from the vendored path.

```
~/Downloads$ gem install --no-rdoc --no-ri --install-dir files/default/vendor --no-user-install inspec
...
...
~/Downloads$ du -sk files
27480	files  # 27MB
~/Downloads$
```

chris-rock:
As discussed with @grdnrio @jeremymv2 @alexpop:

- We identified that at least 90% of the customers do not have their own gem repository
- for now, we added alternative gem source: https://github.com/chef-cookbooks/audit/pull/146
- If we are going to vendor, we should eliminate all non-required dependencies (eg. winrm or docker for train )

chris-rock:
InSpec is part of Chef 13 now. Therefore we do not need to vendor inspec anymore.

PR #165, opened 2016-11-17
add the ability to install inspec as package when specified

vjeffrey:
### Description

@stephenlauck came up with a nice recipe to do this (https://github.com/stephenlauck/audit/blob/lauck/inspec_package/recipes/inspec_package.rb), just filling in the details to get it working with attributes, etc

Adds the ability to install inspec as package when specified.  @stephenlauck does this fit what you were thinking??

### Issues Resolved

fixes: https://github.com/chef-cookbooks/audit/issues/164

### Check List

- [x] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [x] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>


recipes/inspec_package.rb
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+chef_ingredient 'inspec' do
+  version node['audit']['inspec_version'] if node['audit']['inspec_version'] != 'latest'
+  accept_license true
+  action :upgrade
+end
+
+load_inspec_libs
```
jeremymv2: Have you tested this with the package?

The package seems to have `/opt/inspec/embedded/lib` as its base path.  I don't think that's in Chef-client's $LOAD_PATH..

jeremymv2:
I see the value in installing via package.  I am pretty sure though that enterprises are not going to pull the package down from our own repos but will want to host it internally on theirs.  That would need to be a capability in my opinion.

recipes/inspec_package.rb
jeremymv2: That appears to be the case:

```
root@node:/tmp# /opt/chef/embedded/bin/irb
irb(main):001:0> require 'inspec'
=> true
irb(main):002:0>
root@node:/tmp# /opt/chef/embedded/bin/gem which inspec
/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.4.1/lib/inspec.rb
root@node:/tmp# /opt/chef/embedded/bin/gem uninstall inspec
Remove executables:
    inspec

in addition to the gem? [Yn]  y
Removing inspec
Successfully uninstalled inspec-1.4.1
root@node:/tmp# /opt/chef/embedded/bin/irb
irb(main):001:0> require 'inspec'
LoadError: cannot load such file -- inspec
    from /opt/chef/embedded/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /opt/chef/embedded/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from (irb):1
    from /opt/chef/embedded/bin/irb:11:in `<main>'
irb(main):002:0> libpath = '/opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.4.1/lib'
=> "/opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.4.1/lib"
irb(main):003:0> $LOAD_PATH.unshift(libpath) unless $LOAD_PATH.include?(libpath)
=> ["/opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.4.1/lib", "/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/did_you_mean-1.0.0/lib", "/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/rb-readline-0.5.3/lib", "/opt/chef/embedded/lib/ruby/site_ruby/2.3.0", "/opt/chef/embedded/lib/ruby/site_ruby/2.3.0/x86_64-linux", "/opt/chef/embedded/lib/ruby/site_ruby", "/opt/chef/embedded/lib/ruby/vendor_ruby/2.3.0", "/opt/chef/embedded/lib/ruby/vendor_ruby/2.3.0/x86_64-linux", "/opt/chef/embedded/lib/ruby/vendor_ruby", "/opt/chef/embedded/lib/ruby/2.3.0", "/opt/chef/embedded/lib/ruby/2.3.0/x86_64-linux"]
irb(main):004:0> require 'inspec'
=> true
irb(main):005:0>
```

vjeffrey: oh shoot i think i had some leftover gem cruft in there when i tested.  right, ok, so we would need to set the load path like you did there. sounds like there are some other issues too (looking in compliance-support) in that chef-ingredient uses mixlib install gem.  so we could go the remote file and execute route. and then there's the problem of having gem and package installed....we could do a check to get around that i suppose

vjeffrey:
oh, yup -- definitely @jeremymv2 good idea, thank you!

vjeffrey:
ok so things to do here are: 
1) change from using chef_ingredient to remote file execute?
2) make sure path is set
3) ensure there's no gem already installed if doing package install?
4) give ability to specify package source

does that sound right @stephenlauck @jeremymv2 ??

jeremymv2:
Here's a time proven example of chef_ingredient installing supermarket rpm from source or potentially from custom repo: https://github.com/chef-cookbooks/supermarket-omnibus-cookbook/blob/master/resources/supermarket_server.rb

The `node['chef-ingredient']['custom-repo-recipe']`: https://github.com/chef-cookbooks/chef-ingredient#chef_ingredient

stephenlauck:
Would it be possible for the Inspec package install to configure itself and be ready to be loaded so the recipe is only does a package install and Inspec is ready to go?

vjeffrey:
from discussion:  it might be preferable to vendor the gem in the cookbook. 4+ big customers would benefit from a vendored setup

jeremymv2:
Please refer to: https://github.com/chef-cookbooks/audit/issues/112
Vendoring a gem is not without its own caveats.

jeremymv2:
Vendoring a gem and its dependencies in a cookbook is not a pattern that is commonly used.  If hosting the gem is internally is not an option, then this could be an alternative strategy:

You can bundle up the gems you need into a tarball, host that as an artifact internally, then download and extract via recipe in your wrapper cookbook:
```
# example with mixlib-install gem and dependencies
$ gem install --no-rdoc --no-ri --install-dir tmp/ --no-user-install mixlib-install
Fetching: artifactory-2.3.3.gem (100%)
Successfully installed artifactory-2.3.3
Fetching: mixlib-versioning-1.1.0.gem (100%)
Successfully installed mixlib-versioning-1.1.0
Fetching: mixlib-shellout-2.2.6.gem (100%)
Successfully installed mixlib-shellout-2.2.6
Fetching: mixlib-install-1.1.0.gem (100%)
Successfully installed mixlib-install-1.1.0
4 gems installed

$ ls -l tmp/
drwxr-xr-x  2 jmiller  staff   68 Jul 19 19:01 build_info
drwxr-xr-x  6 jmiller  staff  204 Jul 19 19:01 cache
drwxr-xr-x  2 jmiller  staff   68 Jul 19 19:01 doc
drwxr-xr-x  2 jmiller  staff   68 Jul 19 19:01 extensions
drwxr-xr-x  6 jmiller  staff  204 Jul 19 19:02 gems
drwxr-xr-x  6 jmiller  staff  204 Jul 19 19:02 specifications
$ cd tmp
$ tar cvf gems.tar *
Use it later in a recipe:

download_location = ::File.join(Chef::Config[:file_cache_path], 'gems.tar')

remote_file download_location do
  source 'https://your.artifact.server/gems.tar'
  action :create
end

execute 'extract gems' do
  command "tar -xvf #{download_location} -C #{node[:languages][:ruby][:gems_dir]}"
  not_if do
    ::File.exist?("#{node[:languages][:ruby][:gems_dir]}/somegem")
  end
end
```

vjeffrey:
@stephenlauck @jerryaldrichiii

recipes/default.rb
```diff
@@ -17,7 +17,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if node['audit']['inspec_package_source']
+if node['audit']['inspec_package']['source']
   include_recipe 'audit::inspec_package'
```
jeremymv2: Curious if there is a reason to break out inspec installation into two separate recipes?  Would it be tighter to just `include_recipe 'audit::inspec'` here and let the `inspec` recipe have the conditional logic to decide whether to install from package or from gem within an if statement?

spec/unit/recipes/default_spec.rb
```diff
@@ -72,33 +72,26 @@
   context 'When a package install is specified' do
     let(:chef_run) do
       ChefSpec::ServerRunner.new do |node|
-        node.override['audit']['inspec_package'] = true
+        node.override['audit']['inspec_package']['source'] = 'http://path/to/fake.rpm'
```
jeremymv2: I think the inspec package installation tests should be moved into a `inspec_spec.rb` file under `spec/unit/recipes` since the best practice is for the unit test file name to be `<recipe_name>_spec.rb`.

recipes/default.rb
vjeffrey: oh, ya, that makes sense @jeremymv2 i like that.

spec/unit/recipes/default_spec.rb
vjeffrey: :+1:

vjeffrey:
pushed a commit to address your comments @jeremymv2 ; thank you for reviewing. Also went ahead and added info to readme.

jeremymv2:
This doesn't work on Ubuntu since the default package provider selected is `apt_package`

```
           Chef::Exceptions::Package
           -------------------------
           apt package provider cannot handle source attribute. Use dpkg provider instead
```

vjeffrey:
huh....shouldn't the chef package resource be doing that for us? `package tells the chef-client to manage a package; the chef-client will determine the correct package provider to use based on the platform running on the node`

jeremymv2:
The package resource is making a decision, but choosing the wrong one when `source` property is specified.  I suppose the Package Resource should be smart enough to detect that, but that doesn't appear to be the case :(  `apt_package` seems to be the default rather than `dpkg_resource`.

vjeffrey:
k...so doing something along these lines is working for me:
```
  case node[:platform]
  when 'ubuntu'
    dpkg_package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  when 'centos'
    package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  end
```

jeremymv2:
Cool. Need to handle when !ubuntu && !centos
Perhaps just `package` is a sane default for (Windows, Mac, ..)

vjeffrey:
k...so how much is a reasonable scope?  we can make the default one package, as is, which covers centos...then dpkg package for ubuntu, bff package for aix, freebsd package for freebsd, solaris package for solaris, windows package for windows....so we end up with something like:
```
  case node[:platform]
  when 'ubuntu'
    dpkg_package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  when 'aix'
    bff_package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  when 'freebsd'
    freebsd_package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  when 'solaris'
    solaris_package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  when 'windows'
    windows package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  else
    package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  end
```

:/  anything from that list that 'isn't really worth it' or anything that should be added?

jeremymv2:
Browsing the chef-client code base under `chef/lib/chef/provider/package` I'm only seeing where `source` is an issue for `apt_package`.  That being the case, I think this would suffice and handle all scenarios:

```
case node[:platform]
  when 'ubuntu'
    dpkg_package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  else
    package 'inspec' do
      source package_path
      action :nothing
    end.run_action(:install)
  end
```

vjeffrey:
oh good, that's much nicer!

jeremymv2:
Awesome improvement to the cookbook!  👍

chris-rock:
InSpec gem is part of Chef 13, therefore we do not need to work on alternatives anymore.

chris-rock:
Thank you @vjeffrey and @jeremymv2 for raising the issue. This helped us to bring inspec into chef-client!

PR #180, opened 2016-12-07
throw chef-client exception if requested by users

### LINKED ISSUE #166: fail_if_not_present doesn't work
### Cookbook version
2.1.0

### Chef-client version
12.9

### Platform Details
CentOS 7.2, Chef Server 12

### Scenario:
If the compliance profile does not exist and the attribute ['audit']['fail_if_not_present'] is TRUE, the cookbook should failed to converge. This does not happen. Everything converges as normal, but no compliance profiles are run.

### Steps to Reproduce:
Run audit cookbook with fail_if_not_present set to true and a non-existent compliance profile in ['audit']['profiles']

### Expected Result:
Should fail to converge

### Actual Result:
Does not fail to converge

chris-rock:
### Description

This enables user to fail a chef-client run, if a some parts fail during reporting stage

### Issues Resolved

fixes #166 

Thank you @stevendanna for highlighting the solution

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -79,6 +79,32 @@ suites:
           - git: https://github.com/dev-sec/tests-ssh-hardening.git
           - name: ssh
             supermarket: hardening/ssh-hardening
+  # default setting does not fail for missing profile
+  - name: missing-profile-no-fail
+    run_list:
+      - recipe[test_helper::setup]
+      - recipe[audit::default]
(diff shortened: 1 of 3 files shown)


alexpop:
Works a expected, thanks Chris!

chris-rock:
We released audit cookbook 2.3.2. That is including the fix.

PR #181, opened 2016-12-08
Travis and kitchen-ec2 testing

alexpop:
Uses IAM policy that can only create, tag and terminate own ec2 instances


test/kitchen-automate/test/integration/default/chef-server-visibility_spec.rb
```diff
@@ -31,24 +31,33 @@
...
+describe elastic_hits.length do
   it { should eq 2 }
 end
 
-describe elastic_hits[0]['_source']['compliance_summary'] do
-  its(['status']) { should eq 'failed' }
-  its(['total']) { should be > 60 }
-end
+# guard the elastic_hits array from index out of bounds exception
+if elastic_hits.length == 2
```
chris-rock: Should we add an error if the length is not 2?

alexpop: I was thinking that this will provide the error:
```ruby
describe elastic_hits.length do
    it { should eq 2 }
end
```

chris-rock: You are right!

PR #183, opened 2016-12-12
Mention uploading profiles to Automate

alexpop:
Docs improvement after helping a person in the community figure out why profiles could not be downloaded from the profiles API.

User was under the impression that profile `base/windows` was being shipped by default with Chef Automate.


chris-rock:
@alexpop Happy to merge, once the DCO is green

PR #192, opened 2017-01-05
make automate integration tests optional

chris-rock:
Automate integration tests do not run for external PRs. Therefore we should make them optional.


alexpop:
Thanks Chris, taking every opportunity to improve things!

PR #196, opened 2017-02-16
Type the inspec profile attributes

alexpop:
Some document stores like ElasticSearch don't like values that change type. This change converts all profile attribute defaults to string and adds a 'type' key to store the original type.

Some testing improvements.


arlimus:
I like it, thank you Alex!

fyi we had a discussion today on this field. InSpec is doing the right thing with keeping the type intact inside the JSON. however some systems just don't understand how to handle that.

2 workarounds, that keep the type intact, are adding a type field or double-encoding it into a json-string. we choose the former here.

Kudos Alex 👍

Issue #200, opened 2017-03-13
Audit cookbook failing to install from internal Ruby gem mirror


smford22:
### Cookbook version
`2.4.0`

### Chef-client version
`12.19.36`

### Platform Details
OEL 5.11

### Scenario:
Running the `audit::default` recipe on a node fails to install latest `inspec` from an internal mirror

### Steps to Reproduce:
Update `default['audit']['inspec_gem_source']` to point to an internal mirror and converge chef-client

### Expected Result:
The internal mirror has the latest version of inspec staged as well as its dependencies. When `chef-client` runs it should pull inspec and its dependencies from the mirror and install them on the node

### Actual Result:
```
* chef_gem[inspec] action install[2017-03-10T10:13:21-08:00] INFO: Processing chef_gem[inspec] action install (audit::inspec line 20)
[2017-03-10T10:13:36-08:00] INFO: chef_gem[inspec] installed inspec at 1.15.0

   - install version 1.15.0 of package inspec

 ================================================================================
 Recipe Compile Error in /var/chef/cache/cookbooks/audit/recipes/default.rb
 ================================================================================

 Gem::ConflictError
 ------------------
 Unable to activate train-0.22.0, because net-ssh-4.0.1 conflicts with net-ssh (< 4.0, >= 2.9)

 Cookbook Trace:
 ---------------
   /var/chef/cache/cookbooks/audit/libraries/compliance.rb:5:in `load_inspec_libs'
   /var/chef/cache/cookbooks/audit/recipes/inspec.rb:28:in `from_file'
   /var/chef/cache/cookbooks/audit/recipes/default.rb:20:in `from_file'

 Relevant File Content:
 ----------------------
 /var/chef/cache/cookbooks/audit/libraries/compliance.rb:

   1:  # encoding: utf-8
   2:
   3:  # load all the inspec and compliance bundle requirements
   4:  def load_inspec_libs
   5>>   require 'inspec'
   6:    if Inspec::VERSION != node['audit']['inspec_version'] && node['audit']['inspec_version'] !='latest'
   7:      Chef::Log.warn "Wrong version of inspec (#{Inspec::VERSION}), please "\
   8:        'remove old versions (/opt/chef/embedded/bin/gem uninstall inspec).'
   9:    else
  10:      Chef::Log.warn "Using inspec version: (#{Inspec::VERSION})"
  11:    end
  12:    require 'bundles/inspec-compliance/api'
  13:    require 'bundles/inspec-compliance/http'
[...]

PR #203, opened 2017-03-22
Only install InSpec if not installed or version provided

adamleff:
To assist with air-gapped installations, the audit cookbook is
being modified to only install InSpec if it's not already installed
or if the user has explicitly stated a version to be installed.

This will allow users to use other methods of installing the gem
in the chef gem path, or take advantage of InSpec being included
in chef client in the future.

Signed-off-by: Adam Leff <adam@leff.co>

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -17,7 +17,7 @@
 
 # controls inspec gem version to install
 # example values: '1.1.0', 'latest'
-default['audit']['inspec_version'] = '1.15.0'
+default['audit']['inspec_version'] = nil
 
 # sets URI to alternate gem source
 # example values: nil, 'https://mygem.server.com'
--- a/resources/inspec_gem.rb
+++ b/resources/inspec_gem.rb
@@ -1,0 +1,58 @@
+provides :inspec_gem
+resource_name :inspec_gem
+
+property :gem_name, String, name_property: true
+property :version, String
+property :source, String
+
+default_action :install
+
+action :install do
+  if !version.nil?
+    converge_by 'uninstall all inspec and train gem versions' do
+      uninstall_inspec_gem
+    end
+    converge_by "install requested inspec version #{version}" do
+      install_inspec_gem(version: version, source: source)
+    end
+  elsif !inspec_installed?
+    converge_by 'install latest InSpec version' do
+      install_inspec_gem(source: source)
+    end
+  else
+    Chef::Log.info("inspec_gem: not installing InSpec. It's already installed and an explicit version was not supplied.")
+  end
+end
+
+action_class do
+  def install_inspec_gem(options)
+    gem_source  = options[:source]
+    gem_version = options[:version]
+
+    chef_gem 'inspec' do
+      version gem_version if !gem_version.nil? && gem_version != 'latest'
+      clear_sources true unless gem_source.nil?
+      source gem_source unless gem_source.nil?
+      action :install
+    end
+  end
+
+  def uninstall_inspec_gem
+    chef_gem 'remove all inspec versions' do
+      package_name 'inspec'
+      action :remove
+    end
+
+    chef_gem 'remove all train versions' do
+      package_name 'train'
+      action :remove
+    end
+  end
+
+  def inspec_installed?
+    require 'inspec'
+    true
+  rescue LoadError
+    false
+  end
+end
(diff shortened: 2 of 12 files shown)


resources/inspec_gem.rb
```diff
@@ -0,0 +1,41 @@
...
+      install_inspec_gem(nil, source)
+    end
+  else
+    Chef::Log.info("inspec_gem: not installing InSpec. It's already installed and an explicit version was not supplied.")
+  end
+end
+
+action_class do
+  def install_inspec_gem(gem_version, gem_source)
+    chef_gem 'inspec' do
```
chris-rock: Do you think it makes sense to remove inspec and train before reinstall? Especially if a version is set, we may get conflicts, since newer versions of the dependencies are not installed.

attributes/default.rb
```diff
@@ -17,7 +17,7 @@
 
 # controls inspec gem version to install
 # example values: '1.1.0', 'latest'
-default['audit']['inspec_version'] = '1.15.0'
+default['audit']['inspec_version'] = nil
```
chris-rock: we should document that we use latest by default now

resources/inspec_gem.rb
```diff
@@ -0,0 +1,41 @@
...
+resource_name :inspec_gem
+
+property :gem_name, String, name_property: true
+property :version, String
+property :source, String
+
+default_action :install
+
+action :install do
+  if !version.nil?
```
chris-rock: That is a great improvement!

resources/inspec_gem.rb
```diff
@@ -0,0 +1,41 @@
...
+    converge_by "install InSpec version #{version}" do
```
chris-rock: Should we double-check the existing version? Since it tries to install it every time once a version is set.

test/integration/default/default_spec.rb
```diff
@@ -11,7 +11,7 @@
 
 # TODO: change once gem resource handles alternate path to `gem` command
 describe command('/opt/chef/embedded/bin/gem list --local -a -q inspec | grep \'^inspec\' | awk -F"[()]" \'{printf $2}\'') do
```
chris-rock: we should use the new method once its merged to inspec, see https://github.com/chef/inspec/pull/1596

resources/inspec_gem.rb
adamleff: I'm partially against that since we should be trusting chef (and chef_gem) to do what we tell it to. But I'm all for setting us up for greater success... to your next comment about removing-then-installing, I think that's a fine idea and will test that.

resources/inspec_gem.rb
adamleff: Yup, since a user may downgrade an image, etc... I think it's fine to do that. I'll write that in.

attributes/default.rb
adamleff: I'll update the README.

adamleff:
@alexpop I wrote this change in a way that we should hopefully not have to wait until Chef 13, but assuming my Chef change is merged, this should only be better in Chef 13, eliminating most of the gem installs.

I'm fine bumping the major version. The only major change IMO is that we'll install the latest InSpec instead of being hard-locked to 1.15.0. If we feel this is a breaking change, I'm fine with it. However, I'd venture to say hard-locking to 1.15 is semi-bug-worthy :)

chris-rock:
I agree with @adamleff Since inspec is still installed by default, it is not a breaking change. This would also mean, that we remove https://github.com/chef-cookbooks/audit/blob/master/recipes/default.rb#L20 once Chef 13 is out and bump the major version.

adamleff:
I thought about this some more, and I think the change in install behavior (i.e. don't install if it's already installed) could technically be perceived as a breaking change, so I'm going to leave the change I introduced to bump the major version. Upgrading will only help most of our customers, but if anyone was expecting a version other than 1.15.0 to be installed every time, they're going to be caught off-guard.

Bumping the major version still feels like the right thing to do here. I expect most people aren't pinning this cookbook anyways but I'll post to discourse prior to the release.

smurawski:
This would also resolve #200

test/integration/gem-install-version/default_spec.rb
```diff
@@ -0,0 +1,6 @@
+#inspec (1.16.1)
```
alexpop: `v1.16.1` needs updating or just not mention the version in the comments

PR #208, opened 2017-04-25
better error output

chris-rock:
- prints installed InSpec version
- improves the error message if profiles could not be fetched
- removes stacktrace if profile fetching went wrong


arlimus:
Easy, effective, and super helpful thank you!

PR #237, opened 2017-06-08
Ensure min version of inspec is used

alexpop:
With the cookbook no longer specifying an inspec_version in attributes, it's important to ensure a compatible inspec is used.

I'm tempted to bump the required version to at least `1.25.0` because of this: https://github.com/chef/inspec/pull/1816

or should I make it 1.26 or 1.27?

### DIFF
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -124,7 +124,13 @@ class Chef
 
       # run profiles and return report
       def call(opts, profiles)
-        Chef::Log.info "Initialize InSpec #{::Inspec::VERSION}"
+        Chef::Log.info "Using InSpec #{::Inspec::VERSION}"
+
+        if Gem::Version.new(Inspec::VERSION) < Gem::Version.new('1.24.0')
+          Chef::Log.error "This audit cookbook version requires InSpec 1.24.0 or newer, aborting compliance scan..."
+          return {}
+        end
+
         Chef::Log.debug "Options are set to: #{opts}"
         runner = ::Inspec::Runner.new(opts)
 
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -5,6 +5,8 @@ class Chef
   class Handler
     # Creates a compliance audit report
     class AuditReport < ::Chef::Handler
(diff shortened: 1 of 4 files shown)


files/default/handler/audit_report.rb
```diff
@@ -124,7 +124,13 @@ def get_opts(format, quiet)
 
       # run profiles and return report
       def call(opts, profiles)
-        Chef::Log.info "Initialize InSpec #{::Inspec::VERSION}"
+        Chef::Log.info "Using InSpec #{::Inspec::VERSION}"
+
+        if Gem::Version.new(Inspec::VERSION) < Gem::Version.new('1.24.0')
```
adamleff: I'd store this in a constant somewhere near the top since the version string is used in two places: the comparison and the log message. That way we only have to change it in one location.

files/default/handler/audit_report.rb
```diff
@@ -124,7 +124,13 @@ def get_opts(format, quiet)
...
+          Chef::Log.error "This audit cookbook version requires InSpec 1.24.0 or newer, aborting compliance scan..."
+          return {}
```
adamleff: I'd honestly raise this as an exception. People will miss the log message and wonder why they're losing data.

chris-rock:
With Chef 13 we need to work with the version that is shipped with Chef. We may need to shim parts if we really require it here. Otherwise we need to raise the requirement for Chef client version

chris-rock:
Just double-checked all inspec versions in Chef 13:
```
chef 13.0.113: InSpec 1.19.1
chef 13.0.118: InSpec 1.19.1
chef 13.1.31: InSpec 1.25.1
```
Therefore lets use the minimum version of 1.25.1. This allows us to establish a clear path for customers:

- chef 12 + InSpec 1.25.1
- chef 13.1.31

files/default/handler/audit_report.rb
chris-rock: I agree on @adamleff proposal. It is better to fail early

chris-rock:
I am going to bump the requirement to 1.25.1 to ensure everything works with that version. We may need to add some shims later.

chris-rock:
Thank you @alexpop and @adamleff

PR #235, opened 2017-06-07
Update to use v1.27.0 Compliance::API of "inspec" gem.

qubitrenegade:
### Description

[v1.27.0](http://www.rubydoc.info/gems/inspec/1.27.0/Compliance/API#version-class_method) of Compliance::API#version only takes one parameter, this cookbook is using the API of [v1.24.0](http://www.rubydoc.info/gems/inspec/1.24.0/Compliance/API#version-class_method) of Compliance::API#version which used to take two parameters.

Update to use v1.27.0 Compliance::API of `inspec` gem.

### Issues Resolved

#232

### Check List

- [X] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [N/A] New functionality includes testing. (bug fix)
- [N/A] New functionality has been documented in the README if applicable (bug fix)
- [X] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

PR #251, opened 2017-06-29
fix InSpec 1.27.0 Compliance::API use

chris-rock:
rebase of #235


chris-rock:
Thank you @qubitrenegade for this fix

PR #262, opened 2017-07-12
Support for attributes within audit cookbook

### LINKED ISSUE #261: Support inspec attributes
I want to be able to pass in attributes to the inspec profile(s) that are in my list of profiles.

The inspec profile can use [attributes](http://www.anniehedgie.com/inspec-basics-9) and passed in an attributes.yml file. The chef run should be able to define these attributes, per profile, like this:

```json
"audit": {
  "reporter": "chef-server-compliance",
  "inspec_version": "1.29.0",
  "profiles": [
    {
      "name": "linux",
      "compliance": "base/linux",
      "attributes": {
        "role": "mywebserver"
      }
    }
  ]
}
```

The `audit` cookbook will take attributes (if they exist) and pass those to the inspec profile when it runs it.

This enables "one profile to rule them all" pattern that @jeremymv2 proposed, but with a more inspec-native construct.

mhedgpeth:
Fixes #261 

Allows for `node['audit']['attributes']` to define which attributes are passed into the InSpec run
Added documentation in `README.md`
Tested in kitchen, which relies on the [attributes-file-exists-profile](https://github.com/mhedgpeth/attribute-file-exists-profile) for testing that a file exists, driven off of an attribute.

@jeremymv2 this is what we were talking about as a better way to manage the "profile that rules them all" pattern.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -8,6 +8,8 @@ transport:
 
 provisioner:
   name: dokken
+  chef_log_level: info
+  chef_output_format: minimal
 
 verifier:
   name: inspec
@@ -77,11 +79,12 @@ suites:
           - name: ssh-hardening
             url: https://github.com/dev-sec/tests-ssh-hardening/archive/master.zip
           - git: https://github.com/dev-sec/tests-ssh-hardening.git
-          - name: ssh
-            supermarket: hardening/ssh-hardening
+          - name: ssh-baseline
+            supermarket: dev-sec/ssh-baseline
   # default setting does not fail for missing profile
   - name: attributes-work
     run_list:
+      - recipe[test_helper::setup]
       - recipe[test_helper::create_file]
       - recipe[audit::default]
     attributes:
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -72,7 +72,7 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         profiles:
           - name: ssh-hardening
             url: https://github.com/dev-sec/tests-ssh-hardening/archive/master.zip
@@ -86,10 +86,9 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         profiles:
-          - name: attribute-file-exists
-            git: https://github.com/mhedgpeth/attribute-file-exists-profile.git
+          - git: https://github.com/mhedgpeth/attribute-file-exists-profile.git
         attributes:
           file: /opt/kitchen/cache/attribute-file-exists.test
   - name: missing-profile-no-fail
@@ -98,7 +97,7 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         profiles:
           - name: ssh-hardening
             url: https://github.com/dev-sec/this-is-not-available.zip
@@ -110,7 +109,7 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         fail_if_not_present: true
         profiles:
           - name: ssh-hardening
@@ -132,7 +131,7 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         inspec_version: 1.29.0
         fail_if_not_present: true
   - name: skip-inspec-gem-install
@@ -141,7 +140,7 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         inspec_version: 1.25.1
         fail_if_not_present: true
   - name: chef-12-test
@@ -149,6 +148,6 @@ suites:
       - recipe[audit::default]
     attributes:
       audit:
-        collector: json-file
+        reporter: json-file
         inspec_version: 1.25.1
         fail_if_not_present: true
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -80,6 +80,18 @@ suites:
           - name: ssh
             supermarket: hardening/ssh-hardening
   # default setting does not fail for missing profile
+  - name: attributes-work
+    run_list:
+      - recipe[test_helper::create_file]
+      - recipe[audit::default]
+    attributes:
+      audit:
+        collector: json-file
+        profiles:
+          - name: attribute-file-exists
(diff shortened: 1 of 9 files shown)


seththoenen:
+1

Gemfile
```diff
@@ -21,6 +21,8 @@ group :test do
   gem 'rb-readline'
   gem 'webmock'
   gem 'fauxhai', '~> 5.2'
+  gem 'pry-remote'
```
chris-rock: Do we need that for attributes support?

README.md
```diff
@@ -163,6 +163,17 @@ default["audit"] = {
 }
 ```
 
+#### Attributes
+
+You can also pass in [InSpec Attributes](http://www.anniehedgie.com/inspec-basics-10) to your audit run. You do this by defining the attributes here:
```
chris-rock: I would prefer if we could reference to the offical docs https://www.inspec.io/docs/reference/profiles/

test/cookbooks/test_helper/recipes/create_file.rb
```diff
@@ -0,0 +1,5 @@
+# ensures that the file defined by attributes exists, so its associated profile will pass
```
chris-rock: Is this profile missing?

.kitchen.yml
```diff
@@ -80,6 +80,18 @@ suites:
           - name: ssh
             supermarket: hardening/ssh-hardening
   # default setting does not fail for missing profile
+  - name: attributes-work
```
chris-rock: Can we add some integration tests?

files/default/handler/audit_report.rb
```diff
@@ -112,14 +113,15 @@ def load_automate_fetcher
...
+      def get_opts(format, quiet, attributes)
         output = quiet ? ::File::NULL : $stdout
         Chef::Log.warn "Format is #{format}"
         opts = {
           'report' => true,
           'format' => format,
           'output' => output,
-          'logger' => Chef::Log, # Use chef-client log level for inspec run
+          'logger' => Chef::Log, # Use chef-client log level for inspec run,
+          'attributes' => attributes,
```
jeremymv2: The attributes key needs to be a symbol, otherwise it's not working; ie. `attributes: attributes,`

As-is, you can see that there's another symbol key that is not assigned the value I'm assigning:

```
From: /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.30.0/lib/inspec/runner.rb @ line 124 Inspec::Runner#load_attributes:

    120: def load_attributes(options)
    121:   options[:attributes] ||= {}
    122:   require 'pry';binding.pry
    123:
 => 124:   secrets_targets = options[:attrs]
    125:   return options[:attributes] if secrets_targets.nil?
    126:
    127:   secrets_targets.each do |target|
    128:     secrets = Inspec::SecretsBackend.resolve(target)
    129:     if secrets.nil?
    130:       raise Inspec::Exceptions::SecretsBackendNotFound,
    131:             "Unable to find a parser for attributes file #{target}. " \
    132:             'Check to make sure the file exists and has the appropriate extension.'
    133:     end
    134:
    135:     next if secrets.attributes.nil?
    136:     options[:attributes].merge!(secrets.attributes)
    137:   end
    138:
    139:   options[:attributes]
    140: end

[1] pry(#<Inspec::Runner>)> options
=> {"report"=>true,
 "format"=>"json",
 "output"=>"/dev/null",
 "logger"=>Chef::Log,
 "attributes"=>{"password"=>"bob"},
 :logger=>#<Logger:0x00000004f37010 @default_formatter=#<Logger::Formatter:0x00000004f36fc0 @datetime_format=nil>, @formatter=nil, @level=0, @logdev=nil, @progname=nil>,
 :attributes=>{}}
[2] pry(#<Inspec::Runner>)>
```

When line 124 is changed to `attributes: attributes,`:

```
From: /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.30.0/lib/inspec/runner.rb @ line 124 Inspec::Runner#load_attributes:

    120: def load_attributes(options)
    121:   options[:attributes] ||= {}
    122:   require 'pry';binding.pry
    123:
 => 124:   secrets_targets = options[:attrs]
    125:   return options[:attributes] if secrets_targets.nil?
    126:
    127:   secrets_targets.each do |target|
    128:     secrets = Inspec::SecretsBackend.resolve(target)
    129:     if secrets.nil?
    130:       raise Inspec::Exceptions::SecretsBackendNotFound,
    131:             "Unable to find a parser for attributes file #{target}. " \
    132:             'Check to make sure the file exists and has the appropriate extension.'
    133:     end
    134:
    135:     next if secrets.attributes.nil?
    136:     options[:attributes].merge!(secrets.attributes)
    137:   end
    138:
    139:   options[:attributes]
    140: end

[1] pry(#<Inspec::Runner>)> options
=> {"report"=>true,
 "format"=>"json",
 "output"=>"/dev/null",
 "logger"=>Chef::Log,
 :attributes=>{"password"=>"bob"},
 :logger=>#<Logger:0x00000003d4acd0 @default_formatter=#<Logger::Formatter:0x00000003d4ac80 @datetime_format=nil>, @formatter=nil, @level=0, @logdev=nil, @progname=nil>}
[2] pry(#<Inspec::Runner>)>
```

With the change, my inspec control test works as expected, without, I get nil for the attribute value.

As I mentioned, this is because of the key being symbol: https://github.com/chef/inspec/blob/master/lib/inspec/runner.rb#L121

chris-rock:
@mhedgpeth anything I can help with?

test/cookbooks/test_helper/recipes/create_file.rb
adamleff: No, it's in the .kitchen.yml.

test/integration/attributes-work/default.rb
```diff
@@ -0,0 +1 @@
+# will put a test that makes sure a file is there
```
adamleff: Because the profile that actually consumes and uses the attribute is elsewhere (and run by the audit cookbook, not TK), and this is empty, TK won't actually be doing any real testing of this new test suite. To make matters worse, Chef will still exit with a zero exit code even if the audit cookbook failed to scan properly or if there are compliance failures. 🙂 

Can I recommend we add a test here that reads in the JSON outputted by the audit cookbook, and ensures that our test succeeds?

adamleff:
Merging now, will submit a PR to address the issues I raised separately.

chris-rock:
Thank you @mhedgpeth and @adamleff

PR #264, opened 2017-07-27
Change Inspec gem resource action from install to upgrade

jdgoins:
### Description

Changes Inspec gem resource action from `install` to `upgrade`. This allows users to easily reach Inspec version parity on nodes that have an older, incompatible version already installed. 

In other words, the `install` action will not make a change to an existing node's Inspec version which causes the audit to fail, citing a version that is too old. This does not match with the README:
> To install a different version of the InSpec gem, or to force installation of the gem, set the `node['audit']['inspec_version']` attribute to the version you wish to be installed.

### Issues Resolved

n/a

### Check List

- [x] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

### DIFF
--- a/recipes/inspec.rb
+++ b/recipes/inspec.rb
@@ -21,4 +21,4 @@ inspec_gem 'inspec' do
   version node['audit']['inspec_version']
   source node['audit']['inspec_gem_source']
   action :nothing
-end.run_action(:install)
+end.run_action(:upgrade)


adamleff:
@jdgoins thank you for submitting this PR.

Using the `:install` action is on purpose to ensure we don't blindly attempt to upgrade if no version is set. We have many customers that have no internet access and no gem mirrors and therefore rely on the version of InSpec packaged in Chef 13 Omnibus packages. Changing this to upgrade will essentially ignore the `version` directive and cause failures for those users.

Setting the attribute to the desired version is the supported path to getting a newer InSpec gem installed if the one included in the Chef omnibus package is not new enough for your use case.

jdgoins:
@adamleff Thanks for your response, that makes sense. 

It's possible I did something wrong, but the installation of a newer Inspec gem, by setting node['audit']['inspec_version'], did not work for me. I was using w2k12r2 and w2k16 and set `node['audit']['inspec_version'] = '1.31.1'` for existing nodes bootstrapped with chef client 12.19.36. I had to create my own gem_package resource that uses upgrade to get around this.

adamleff:
I'd be interested in seeing the Chef Client run output so we can diagnose. What version of the audit cookbook did this occur in for you?

adamleff:
Closing this PR since the `:install` action is intended to avoid unintended upgrades, and the ability to choose a specific InSpec version still works.

PR #265, opened 2017-08-08
Disable default source when using user-supplied gem source

adamleff:
The `clear_sources` property does not remove the default gem source when the RubyGems package provider does its work. We need to also set the `include_default_source` property to false to ensure users' requesting gem source is the only gem source used (instead of rubygems.org, for example).

### DIFF
--- a/resources/inspec_gem.rb
+++ b/resources/inspec_gem.rb
@@ -42,8 +42,11 @@ action_class do
 
     chef_gem 'inspec' do
       version gem_version if !gem_version.nil? && gem_version != 'latest'
-      clear_sources true unless gem_source.nil?
-      source gem_source unless gem_source.nil?
+      unless gem_source.nil?
+        clear_sources true
+        include_default_source false
+        source gem_source
+      end
       action :install
     end
   end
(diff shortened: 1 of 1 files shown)


jeremymv2:
🤔  the `include_default_source` is only a recent addition to chef-client:
https://github.com/chef/chef/commit/28be4df7aadea1e6f40bc9ea630b2efa30f7639b

adamleff:
@jeremymv2 true, so this would be a question for @chris-rock how far back our backward-compat policy is for the audit cookbook and Chef version.

I'll add some guards around it in the meantime.

adamleff:
For posterity, this property has been in Chef stable since 13.0

jeremymv2:
In the past we've decided on minimum version 12.5.1

https://github.com/chef-cookbooks/audit/blob/master/metadata.rb#L13

I think the guards are the way to go.

adamleff:
And it is now appropriate guarded, @jeremymv2 :)

adamleff:
The Travis tests are due to a breakage between berkshelf/solve and molinillo which cases cookbooks in the dependency chain to go missing. I've tested everything locally using an older ChefDK and it tests well, so I'm going to merge this.

PR #266, opened 2017-08-10
Add test for InSpec Attributes functionality

adamleff:
The work introduced by #262 was lacking an integration test for Test Kitchen. I also renamed the suite to be clearer about what it does and to avoid any confusion that the work may be a work-in-progress.


adamleff:
Travis failures are due to berkshelf/solve and molinillo issues, for which a fix is in the works. Merging.

PR #271, opened 2017-08-29
Pass Chef node to InSpec as an attribute

### LINKED ISSUE #268: Feature enhancement request: Audit cookbook 4.2 to pass node data to Inspec
### Cookbook version
[Version of the cookbook where you are encountering the issue]

### Chef-client version
[Version of chef-client in your environment]

### Platform Details
[Operating system distribution and release version. Cloud provider if running in the cloud]

### Scenario:
[What you are trying to achieve and you can't?]

### Steps to Reproduce:
[If you are filing an issue what are the things we need to do in order to repro your problem? How are you using this cookbook or any resources it includes?]

### Expected Result:
[What are you expecting to happen as the consequence of above reproduction steps?]

### Actual Result:
[What actually happens after the reproduction steps? Include the error output or a link to a gist if possible.]

adamleff:
Providing the Chef node object (attributes, environment, etc.) to InSpec as an attribute will allow the ability to write more flexible profiles and eliminate the need for users to write out JSON files during the converge phase only to get read in by the profile later.

Fixes #268

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -91,6 +91,7 @@ suites:
         reporter: json-file
         profiles:
           - git: https://github.com/mhedgpeth/attribute-file-exists-profile.git
+          - git: https://github.com/adamleff/inspec-profile-chef-node-attributes.git
         attributes:
           file: /opt/kitchen/cache/attribute-file-exists.test
   - name: missing-profile-no-fail
--- a/README.md
+++ b/README.md
@@ -473,6 +473,31 @@ default['audit']['inspec_gem_source'] = 'http://internal.gem.server.com/gems'
 
 Please note that all dependencies to the `inspec` gem must also be hosted in this location.
 
+## Using Chef node data
+
+While it is recommended that InSpec profiles should be self-contained and not rely on external data unless
+necessary, there are valid use cases where a profile's test may exhibit different behavior depending on
+aspects of the node under test.
+
+To assist with these use cases, the audit cookbook will pass the Chef node object as an InSpec attribute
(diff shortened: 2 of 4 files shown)


adamleff:
Travis failures in the "default" suites are unrelated to this change, fixed in #270

mhedgpeth:
@adamleff doesn't this get done by passing in inspec attributes through the cookbook? (My PR?)

mhedgpeth:
@adamleff I believe that designing inspec profiles with a good interface of what **it** needs is the better way. I don't believe that providing the whole node attribute is the right way. Have chef tell inspec what it needs via the attribute feature. Make that as small as possible. That way you can run profiles without an implicit dependency on Chef.

adamleff:
@mhedgpeth your PR is similar to this but not exactly. Your PR allows for certain node attributes to be set (in the 'audit' tree) and then passed as-is. We've seen a number of use cases where people wanted more information, including other non-node information such as run_list, environment, etc. Those individuals were using Chef to write out node objects as JSON and then loading that data in later... and there are a whole suite of reasons why that's a bad practice. This simply provides a better pipe to plumb that data into the InSpec run.

InSpec can't tell Chef what is needs, therefore Chef selectively pass on data to InSpec for what a profile needs without an additional level of coordination. The implicit dependency is only there if the user elects for it to be there.

I agree that having a separation of concerns is better. This in no way ties InSpec to Chef or Chef to InSpec any more than writing out data files and reading them in later. The use of this additional data is optional and we're providing it in an effort to stop the bad practices of littering filesystems with node object JSON files and also meet the needs of many community members who have raised this as a feature request.

PR #274, opened 2017-08-30
Make chef_node attribute an opt-in feature

adamleff:
To encourage good practices, and to not send data to the InSpec runner that is not needed by most users, the `chef_node` attribute that contains the Chef node object hash will only be sent if the "chef_node_attribute_enabled" audit attribute is set to `true`

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -91,9 +91,25 @@ suites:
         reporter: json-file
         profiles:
           - git: https://github.com/mhedgpeth/attribute-file-exists-profile.git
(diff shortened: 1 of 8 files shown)


adamleff:
While it did not make it into the code, let it be known for posterity reasons that the `if` statement on the following line is referred to as "The Hedgpeth `if` Statement":

https://github.com/chef-cookbooks/audit/commit/9955e3b20ab0a141caf312117fd7ac0431e4e30f#diff-ac3f30b2c66fd1ec1b818a9707a4066dR36

Thank you @mhedgpeth for being passionate about the community doing "the right thing." 😄

PR #292, opened 2017-11-06
Send end_time as utc RFC3339

alexpop:
Signed-off-by: Alex Pop <apop@chef.io>


alexpop:
Yes, tested that. We time parse it as `time.RFC3339` and it works fine with both
 `2016-07-19T19:19:19+01:00` and `2016-07-19T18:19:19Z` formats

Issue #296, opened 2017-12-01
Activate inspec cache by default to boost Windows execution


chris-rock:
### Cookbook version

5.0.4

### Chef-client version

n/a

### Platform Details

n/a

### Scenario:

At this point, windows execution is very slow with the audit cookbook. This will be addressed with chef/inspec#2339 and chef/train#220. While the pipes are used by default, command caching is not activated by default. We should activate this `backend_cache=true` flag as a default and allow users to opt-out if required.

chris-rock:
```
"audit": {
  "reporter": "chef-server-automate",
  "fetcher": "chef-server",
  "inspec_backend_cache": true,
  "insecure": false,
  "profiles": [
    {
      "name": "my-profile",
      "compliance": "john/my-profile"
    }
  ]
}
```

PR #297, opened 2017-12-05
Enable Inspec caching

jquick:
### Description

Add inspec caching attribute which defaults to true. This enables the new inspec caching when used by the Audit cookbook.

### Issues Resolved

Fixes https://github.com/chef-cookbooks/audit/issues/296

Signed-off-by: Jared Quick <jquick@chef.io>

### DIFF
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -38,6 +38,11 @@ class Chef
         # load inspec, supermarket bundle and compliance bundle
         load_needed_dependencies
 
+        # check if we have a valid version for backend caching
+        if node['audit']['inspec_backend_cache'] && (Gem::Version.new(::Inspec::VERSION) < Gem::Version.new('1.47.0'))
+          Chef::Log.warn 'inspec_backend_cache requires Inspec version >= 1.47.0'
+        end
+
         # detect if we run in a chef client with chef server
         load_chef_fetcher if reporters.include?('chef-server') ||
                              reporters.include?('chef-server-compliance') ||
@@ -124,6 +129,7 @@ class Chef
           'format' => format,
           'output' => output,
           'logger' => Chef::Log, # Use chef-client log level for inspec run,
+          :backend_cache => node['audit']['inspec_backend_cache'],
           attributes: attributes,
         }
         opts
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -38,10 +38,8 @@ class Chef
         # load inspec, supermarket bundle and compliance bundle
         load_needed_dependencies
 
-        # check if we have a valid version for backend caching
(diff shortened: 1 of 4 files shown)


files/default/handler/audit_report.rb
```diff
@@ -124,6 +129,7 @@ def get_opts(format, quiet, attributes)
           'format' => format,
           'output' => output,
           'logger' => Chef::Log, # Use chef-client log level for inspec run,
+          :backend_cache => node['audit']['inspec_backend_cache'],
```
adamleff: Can we change this to use the new hash syntax instead of hash-rockets?

```ruby
backend_cache: node['audit']['inspec_backend_cache']
```

... the stuff above this apparently needs to be strings (but maybe it doesn't, we should probably check on that as some point), but if this key is expected to be a symbol, we should be consistent. 🙂

files/default/handler/audit_report.rb
```diff
@@ -38,6 +38,11 @@ def report
         # load inspec, supermarket bundle and compliance bundle
         load_needed_dependencies
 
+        # check if we have a valid version for backend caching
+        if node['audit']['inspec_backend_cache'] && (Gem::Version.new(::Inspec::VERSION) < Gem::Version.new('1.47.0'))
```
adamleff: We're now checking for a specific version constraint in two different places within this file... here, and:

https://github.com/chef-cookbooks/audit/blob/master/files/default/handler/audit_report.rb#L134-L137

I think we should move the MIN_INSPEC_VERSION one to be right above this, and out of the `#call` method. Or, possibly better yet, extract both to a `validate_inspec_version!` method and call it here in the `#report` method in place of this stanza. What do you think?

files/default/handler/audit_report.rb
```diff
@@ -38,6 +38,11 @@ def report
...
+          Chef::Log.warn 'inspec_backend_cache requires Inspec version >= 1.47.0'
```
adamleff: `InSpec` - capital S, please :)

spec/unit/report/audit_report_spec.rb
```diff
@@ -58,6 +58,13 @@
   end
 
   describe 'get_opts method' do
+    let(:mynode) { Chef::Node.new }
+
+    before :each do
+      mynode.default['audit']['inspec_backend_cache'] = true
```
adamleff: Should we add a test that assures that `opts[:backend_cache]` gets set to false if the node attribute calls for it? I hate to only have a test for the default path.

metadata.rb
```diff
@@ -5,7 +5,7 @@
 license 'Apache-2.0'
 description 'Allows for fetching and executing compliance profiles, and reporting its results'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version '5.0.4'
+version '5.0.5'
```
adamleff: Gotta be a major bump for a change like this.

jquick:
These issues have been resolved.

PR #328, opened 2018-08-13
Add support for node['audit']['profiles'] as a hash of hashes 

mattray:
Policyfiles with includes do not allow merging arrays and it's a bit unclear why an array was chosen. This patch does not change the existing behavior, only supplements it.

Signed-off-by: Matt Ray <matthewhray@gmail.com>

### DIFF
--- a/.kitchen.vagrant.yml
+++ b/.kitchen.vagrant.yml
@@ -2,22 +2,42 @@
 driver:
   name: vagrant
 
+transport:
+  name: ssh
+
 provisioner:
-  name: chef_solo
+  name: chef_zero
 
 verifier:
   name: inspec
   sudo: true
 
 platforms:
-  - name: centos-7.1
-  - name: centos-6.7
-  - name: centos-5.11
+  - name: centos-6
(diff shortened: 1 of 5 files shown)


chris-rock:
@mattray Great work. Should we update the docs to make the hash map the default now? I like the simpler style.

mattray:
Glad you like it, definitely needed it for Policyfiles. We could add deprecation warnings for array-style usage for a few releases just so we don't blindside anyone.

chris-rock:
I agree a deprecation warning would be great. I cannot recall why we went with arrays in the first place, but maps look very nice

mattray:
I've updated the PR with a deprecation warning.

jblaine:
Note https://github.com/chef-cookbooks/audit/issues/339 where hash of hashes, the new hotness being shown as the replacement for the other 2 methods... doesn't actually work.

mattray:
@jblaine here's a working example: 
https://github.com/mattray/home-repo/blob/master/policyfiles/base.rb#L23

PR #336, opened 2018-09-18
Switch format to reporter

alexpop:
### Description

In prep for the inspec 3.0 release which deprecates format.

### DIFF
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -143,12 +143,13 @@ class Chef
         require 'chef-automate/fetcher'
       end
 
-      def get_opts(format, quiet, attributes)
+      def get_opts(reporter, quiet, attributes)
         output = quiet ? ::File::NULL : $stdout
-        Chef::Log.debug "Format is #{format}"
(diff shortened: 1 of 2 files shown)


files/default/handler/audit_report.rb
```diff
@@ -143,12 +143,13 @@ def load_automate_fetcher
...
 
-      def get_opts(format, quiet, attributes)
+      def get_opts(reporter, quiet, attributes)
         output = quiet ? ::File::NULL : $stdout
-        Chef::Log.debug "Format is #{format}"
+        Chef::Log.debug "Reporter is [#{reporter}]"
         opts = {
           'report' => true,
-          'format' => format,
+          'format' => reporter,      # For compatibility with older versions of inspec. This line can be removed from Q2 2019
```
jerryaldrichiii: On the InSpec side of these we try to phrase these as `TODO:` so we can search for it easily.

Could we do that here or open an issue? I won't hold up the PR over it.

alexpop:
`TODO` makes sense Jerry, pushed another commit to add it.
Thank you for the reviews.

PR #340, opened 2018-10-11
Fix errant dash instead of underscore in example of InSpec version 

gsreynolds:
InSpec version attribute should have an underscore, not a dash, in the README.

Signed-off-by: Gavin Reynolds <gavin@chef.io>


btm:
travis failure is unrelated.

Issue #339, opened 2018-09-28
New Hash of Hashs format for specifying profiles does not work


jquick:
## Issue Description
  New Hash of Hashs format for specifying profiles for the audit cookbook causes an error if actually used.

*Issue Summary*:  In the README.md for the Audit cookbook there is a section that shows examples for specifying profiles to run using a Hash of Hashes (https://github.com/chef-cookbooks/audit#configure-node second set of examples).  

Unfortunately because the "attributes/default.rb" defaults to create the attributes with 

```
default['audit']['profiles'] = []
```

this causes an "no implicit conversion of String into Integer" error to be thrown when attempting to run the wrapper cookbook.

*Expected Results*:  Specifying the profiles using the hash of hashes format should converge successfully without any errors

*Actual Results*:  Converge errors with a "no implicit conversion of String into Integer" TypeError.

*Repro Steps (if possible)*:

1. Create a wrapper cookbook for the Audit cookbook &gt;= 7.1.0 and specify a profile using the following example of the Hash of Hashes format in the README.md.

```
# Hash of hashes, works with Policyfile includes
default['audit']['profiles']['linux'] = { 'compliance': 'base/linux' }
```

2. Converge the cookbook and watch for the error.

*Request for Developers*:  Product Fix

Full exception from kitchen converge

```

       ================================================================================

       Recipe Compile Error in /tmp/kitchen/cache/cookbooks/test/recipes/default.rb

       ================================================================================
   TypeError
   ---------
   no implicit conversion of String into Integer

   Cookbook Trace:
   ---------------
     /tmp/kitchen/cache/cookbooks/test/recipes/default.rb:29:in `from_file'

   Relevant File Content:
   ----------------------
   /tmp/kitchen/cache/cookbooks/test/recipes/default.rb:

    22:  end
    23:
    24:  # node.default['audit']['interval']['enabled'] = true
    25:  node.default['audit']['reporter'] = 'chef-server-automate'
    26:  node.default['audit']['fetcher'] = 'chef-server'
    27:  node.default['audit']['insecure'] = true
    28:
    29&gt;&gt; node.default['audit']['profiles']['DevSec SSH Baseline'] = { compliance: 'admin/ssh-baseline' }
    30:
    31:  include_recipe 'audit::default'
    32:

```

Vasu1105:
[...]

PR #343, opened 2018-11-02
Update profiles default attribute to hash

jquick:
Signed-off-by: Jared Quick <jquick@chef.io>

### Description

When using a wrapper cookbook and trying to add to the profiles attribute using the new hash syntax it fails. This is because the base attribute is still set to an array instead of hash.

### Issues Resolved

This fixes  https://github.com/chef-cookbooks/audit/issues/339

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -85,7 +85,7 @@ default['audit']['overwrite'] = true
 
 # Chef Inspec Compliance profiles to be used for scan of node
 # See README.md for details
-default['audit']['profiles'] = []
+default['audit']['profiles'] = {}
 
 # Attributes used to run the given profiles
 default['audit']['attributes'] = {}


alexpop:
A

jerryaldrichiii:
By forcing a Hash here you prevent the wrapper cookbook(s) from using `.push`. I think it is fine...but should we change the other example?

jquick:
Good point @jerryaldrichiii for some reason I thought that was a valid method on both. This would likely break a decent amount of legacy things. We have had the deprecation warning out for awhile though.

I do think our options are to:
1) Update documentation and make it a breaking change
2) Add `push` to the hash class which alias merge
3) something else

alexpop:
Change  done with https://github.com/chef-cookbooks/audit/pull/385
Thank you!

PR #347, opened 2019-01-09
Audit profiles array deprecation support

ank-s2k:
### Description

Patch to support both array of hashes & hash of hashes profiles format.
- Added deprecations warning if profiles added by re-assign with array of hashes or push methods.
- Extended default profiles object, so that it will support new format(hash of hashes).
- Constructed final array of hashes while adding profiles so that no need to parse the profiles in audit report.

### Issues Resolved

Fixes: https://github.com/chef-cookbooks/audit/issues/339

### Check List

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -87,8 +87,9 @@ default['audit']['overwrite'] = true
 # See README.md for details
 default['audit']['profiles'] = []
 
-# Extend the object and profile capability to support symbol assignment.
-# @TODO: Please remove these extend methods and change default['audit']['profiles'] = {} once we fully support hash of hashes profiles
+# 'default['audit']' and 'default['audit']['profiles']' capabilities are extended to support symbol assignment.
+# @TODO: Once we fully support hash of hashes, then we may remove these extended methods and revert back to using:
+# default['audit']['profiles'] = {}
 default['audit'].extend(Deprecations::Audit::HashProfile)
 default['audit']['profiles'].extend(Deprecations::Audit::ArrayProfile)
 
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -87,6 +87,11 @@ default['audit']['overwrite'] = true
 # See README.md for details
 default['audit']['profiles'] = []
 
+# Extend the object and profile capability to support symbol assignment.
+# @TODO: Please remove these extend methods and change default['audit']['profiles'] = {} once we fully support hash of hashes profiles
+default['audit'].extend(Deprecations::Audit::HashProfile)
+default['audit']['profiles'].extend(Deprecations::Audit::ArrayProfile)
+
 # Attributes used to run the given profiles
 default['audit']['attributes'] = {}
 
--- a/libraries/deprecations.rb
+++ b/libraries/deprecations.rb
@@ -1,0 +1,62 @@
+# encoding: utf-8
+# This module is use to support existing array as well as hash of hashes format
+# that are being assigned by end user.
+#
+# In order to provide backword compativility this module extended by profiles key object
+#
+module Deprecations
+  module Audit
+    # ArrayProfile module extend the hash instance object node['audit']['profiles']
+    # to add warning & format the array assignment.
+    module ArrayProfile
+      # @overload []=(*args)
+      # This method provide the support to assign item as hash(key as String|Symbole)
+      #
+      # @param [String|Symbole] key assigned add as name.
+      # @param [Hash] profile attributes the are being assigned.
+      #
+      # @example format the profile assignment if it is hash of hashes format.
+      #
+      #    node.default['audit']['profiles']['linux'] = { 'compalince': 'base/linux' }
+      #
+      # @example Formatted response return
+      #    { name: linux, 'compalince': 'base/linux' }
+      #
+      # @return [Array<Profiles>]
+      def []=(*args)
+        self << { name: args.first }.merge(args.last.to_h)
+      end
+
+      # @overload push(*args)
+      # WARN msg for the deprication use of array assigment.
+      # @param [Array] Profiles the are being assigned.
+      #
+      # @return [Array<Profiles>]
+      def push(*args)
+        Chef::Log.warn "Use of a hash array for the node['audit']['profiles'] is deprecated. Please refer to the README and use a hash of hashes."
+        super(*args)
+      end
+    end
+
+    # HashProfile module extend the hash instance object node['audit'] to track
+    # re-assignment of the profiles.
+    module HashProfile
+      # @overload []=(key, val)
+      # WARN msg for the deprication use of array assigment.
+      # @param [String|Symbole] key of audit hash.
+      # @param [Array] profiles the are being re-assigned.
+      #
+      # @return [Hash, #key] return hash for non 'profiles' (key, value) pair assign
+      #
+      def []=(key, val)
+        if key.eql?('profiles')
+          Chef::Log.warn "Use of a hash array for the node['audit']['profiles'] is deprecated. Please refer to the README and use a hash of hashes."
+          store(key, val)
+          self[key].extend(ArrayProfile)
+        else
+          super(key, val)
+        end
+      end
+    end
+  end
+end
--- a/libraries/deprecations.rb
+++ b/libraries/deprecations.rb
@@ -32,7 +32,7 @@ module Deprecations
         if key.is_a?(Integer)
           Chef::Log.warn "Use of a hash array for the node['audit']['profiles'] is deprecated. Please refer to the README and use a hash of hashes."
         else
-          index = self.length
+          index = length
           value = { 'name': key }.merge(Hash.try_convert(value) || {})
         end
         super(index, value)
@@ -62,15 +62,18 @@ module Deprecations
         if key.eql?('profiles')
           unless val.is_a?(Hash)
             Chef::Log.warn "Use of a hash array for the node['audit']['profiles'] is deprecated. Please refer to the README and use a hash of hashes."
+            store(key, val)
             self[key].extend(ArrayProfile)
+          else
+            store(key, val)
           end
-          store(key, val)
         else
           super(key, val)
         end
       end
 
       private
+
       # @param key<Object> The key to convert.
       #
       # @param [Object]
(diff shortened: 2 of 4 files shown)


vsingh-msys:
@jerryaldrichiii  Not exactly :smile: Thanks
Could you please also have a look at https://github.com/chef-cookbooks/audit/pull/349?

clintoncwolfe:
I'm pretty surprised at the approach used here - extending the individual attribute elements to be a special class that handles arrays or hashes.

Could someone walk through why this unusual approach was the only option?

vsingh-msys:
Hi @clintoncwolfe,
As per requirement, we need to support both(Array as well as Hash) format of audit profiles.
```
default['audit']['profiles'].push(
    # Profile from Chef Compliance
    {
      'name': 'linux',
      'compliance': 'base/linux'
    }
)
```
And
```
default['audit']['profiles']['linux'] = { 'compliance': 'base/linux' }
```
So In order to support both the format, we need to extend the exsiting Array profiles to support `String|Symbol` indexes(Hash).

Also once we discontinue with profiles assgiment as array we will remove these exteded classes.

Let us know your thought?

attributes/default.rb
```diff
@@ -87,6 +87,12 @@
 # See README.md for details
 default['audit']['profiles'] = []
 
+# 'default['audit']' and 'default['audit']['profiles']' capabilities are extended to support symbol assignment.
+# @TODO: Once we fully support hash of hashes, then we may remove these extended methods and revert back to using:
+# default['audit']['profiles'] = {}
+default['audit'].extend(Deprecations::Audit::HashProfile)
+default['audit']['profiles'].extend(Deprecations::Audit::ArrayProfile)
```
lamont-granquist: There's a bunch of different edge cases here that I can think of around this implementation, but instead of tracking those down what I'd suggest is to abandon this node attribute and pick a new name.  I'm going to call it `default['audit']['profiles-ng']` which is poor and should not be used.  But I can't think of anything better right this second.

Then in the current major version of the cookbook `default['audit']['profiles']` should throw a deprecation warning if it is set to anything other than an empty array.  If `default['audit']['profiles-ng']` is set to a non-empty hash then it should be used instead (there should be no merging and both-set should throw a warning that the old array setting is not being used.

Then in the next major version  of the cookbook `default['audit']['profiles']`  should just throw and fail if its set to anything other than an empty array, and only `default['audit']['profiles-ng']` should be set.

vsingh-msys: @lamont-granquist thanks for the suggestion to store audit profiles within new variable `default['audit']['profiles-ng']`. Also we need minor refactor here https://github.com/chef-cookbooks/audit#configure-node

vsingh-msys: Hey @lamont-granquist,

I like this approach, this is going to be a much simple solution to implement such deprecations.
But about the `default['audit']['profiles']` which are `already set for a node on the server and are not provided by the user while new chef-client run`, could you please let me know if I'm missing something while covering this crucial case:
> (there should be no merging and both-set should throw a warning that the old array setting is not being used.

We should provide backward compatibility to these preloaded `default['audit']['profiles']`, and also **load(merge) them along with `default['audit']['profiles-ng']`** followed by deletion of `default['audit']['profiles']`. later release will throw errors and fail if `default['audit']['profiles']` is set.

vsingh-msys:
Hey @lamont-granquist,

I like this approach, this is going to be a much simple solution to implement such deprecations.
But about the `default['audit']['profiles']` which are `already set for a node on the server and are not provided by the user while new chef-client run`, could you please let me know if I'm missing something while covering this crucial case:
> (there should be no merging and both-set should throw a warning that the old array setting is not being used.

We should provide backward compatibility to these preloaded `default['audit']['profiles']`, and also **load(merge) them along with `default['audit']['profiles-ng']`** followed by deletion of `default['audit']['profiles']`. later release will throw errors and fail if `default['audit']['profiles']` is set.

lamont-granquist:
I think @clintoncwolfe should make the call.  My impression is that this isn't highly used, so providing merging backcompat is not an additional requirement.

attributes/default.rb
jerryaldrichiii: I'm going to update the docs to not mention the deprecated method.

libraries/deprecations.rb
```diff
@@ -0,0 +1,62 @@
...
+      def []=(*args)
+        self << { name: args.first }.merge(Hash.try_convert(args.last) || {})
+      end
+
+      # @overload push(*args)
+      # WARN msg for the use of array assigment.
+      # @param [Array] Profiles the are being assigned.
+      #
+      # @return [Array<Profiles>]
+      def push(*args)
```
btm: There are more Array methods that are missing, such as `<<` that concern us that this will be buggy even if it is temporary.

lamont-granquist: Yeah so the fact that this is being widely used and has a large blast radius to every single user, while it makes going the extra mile to make backcompat work also means that you have to assume that people will be using entirely arbitrary array mutation methods on this attribute.

We do have a complete list of those in Chef::Node::Mixin::ImmutablizeArray::DISALLOWED_MUTATOR_METHODS and a complete list of Hash mutators in Chef::Node::Mixin::ImmutablizeHash::DISALLOWED_MUTATOR_METHODS.

The ones that are specific to Arrays and not Hashes could be included here.  What would we do about the ones that are common across Hashes and Arrays though?  Can't determine if the user intends to be coming from Hash-space or Array-space in that case.  I suppose we could always assume that we need backcompat first and use Array space until the compat layer is removed.

libraries/deprecations.rb
```diff
@@ -0,0 +1,62 @@
...
+    module HashProfile
+      # @overload []=(key, val)
+      # WARN msg for the use of array assigment.
+      # @param [String|Symbol] key of audit hash.
+      # @param [Array] profiles the are being re-assigned.
+      #
+      # @return [Hash, #key] return hash for non 'profiles' (key, value) pair assign
+      #
+      def []=(key, val)
+        if key.eql?('profiles')
```
lamont-granquist: this would also need to handle `:profiles` or just call `convert_key`

libraries/deprecations.rb
```diff
@@ -0,0 +1,62 @@
...
+          Chef::Log.warn "Use of a hash array for the node['audit']['profiles'] is deprecated. Please refer to the README and use a hash of hashes."
```
lamont-granquist: this will also warn on `default['audit']['profiles'] = { ... a valid hash of hashes ... }` so needs to check if the value is an array or a hash.

lamont-granquist:
So I keep thinking this over, and still what we're trying to do here is construct an object which is both Array-like and Hash-like where we attempt to determine which one it should be by the pattern of access by the user which is, in a strong logical sense, non-sensical in ruby.

I don't think we can do this.

It is made actually worse by the fact that the blast radius is everyone that uses the cookbook.

What needs to happen is this needs to follow the pattern of using another attribute (good old filesystem-vs-filesystem2).

We do not need to drop support for Hashes coming from policyfiles.  Also users can go ahead and just blow away the default Array themselves manually and continue to use it.

But for users that need to 'convert' from Arrays to Hashes I think they need to use a `profiles2` object, then at some point in the future a major version drops which drops Array support from `profiles` and then eventually `profiles2` can be dropped.  But that's the only way to run Arrays and Hashes side-by-side coherently.

The other alternative is to just drop a new major version and as part of upgrading to the new major version that attribute must be fixed by the user or they're blocked on the upgrade.  The N-1 is the Array and N+1 is the Hash and there's no dancing around profiles-to-profiles2-and-back.

It also may just be better to document that users should take the upgrade path of manually blowing away the Array with a Hash in their wrapper cookbooks and just do a docs upgrade.

Actually I kinda like that last one.  Fix the docs.

Don't do this.

lamont-granquist:
Okay so the best way I can think of to actually make this work is to more or less leave the `HashProfile` mixin there, so that object hash special behavior so that if I user writes `default['audit']['profiles'] = []` or `default[:audit][:profiles] = []` then we need to give a warning.

But then we need to initialize `default[:audit][:profiles]` earlier with a decorator object.  And really that object needs to make an early decision about if it is really an Array or a Hash -- then it needs to use the `node.__path` information in order to actually rewrite itself out of existence in the node object and replace itself with an actual ImmutableArray and/or ImmutableMash object.

This can still be attacked any number of ways though since before you have an access that can clearly delineate if the object is an Array or a Hash what do you do if someone calls `.is_a?(Hash)` on the object?  It could be, we don't know yet...

vsingh-msys:
Hi @lamont-granquist  @btm,
Thanks for the valuable feedback and will try to cover all edge cases and let you know.

Actually, I was under the impression to provide support only to those methods which are given in the README docs ie. `.push`.

libraries/deprecations.rb
vsingh-msys: Added `convert_key` and called `convert_key` to handle both `'profiles'` and `:profiles`. 
Thanks

libraries/deprecations.rb
vsingh-msys: Added warning for all  `Array` mutator methods.

For the `Hash` and common mutator methods does not do anything because changing the Hash methods behaviour leads to confusion after the array of hashes support will be deprecated.

For example if we provide `.merge` method support to profiles it return  `Array of hashes` instead of `Hash of hashes`.

Please let me know your thoughts?

vsingh-msys:
Hi @lamont-granquist & @btm 
sorry for the late reply!

Tried to catch some of the edges cases and added a warning for the all Array mutator methods.

Could you please have a look?

Thanks

btm:
Rather than the complexity here of trying to support both arrays and hashes, I believe what we want to do is the clean break of making a new major release of the audit cookbook that requires only using hashes. Users will have to switch to hashes when they upgrade.

We'd like to have a less abrupt migration path but I don't think anyone is comfortable with this complexity.

alexpop:
Thank you for the contributions. v9 of the cookbook uses profile hash by default based on this change:
https://github.com/chef-cookbooks/audit/pull/385

Ok to close this PR?

vsingh-msys:
Yup @alexpop go ahead and close it. thanks

alexpop:
Appreciate the quick reply Vivek. Closing. Thank you!

PR #355, opened 2019-02-05
Update test deps and the kitchen dokken config

tas50:
Aligns this with some of our other projects.


kekaichinose:
@miah could I get your eyes on this one? I saw you approved it, and it looks like it only failed on Travis. Do we need to rerun this with BK now?

PR #357, opened 2019-02-13
[WIP] Initial POC of using custom_resources [WIP]

teknofire:
# WORK IN PROGRESS DO NOT MERGE 

### Description

The idea here is to replace the attribute driven nature of this cookbook with custom resources.  

One primary driver for this is to make it easier to handle the transition between the profile list being an Array of Hashes vs. a Hash of Hashes (https://github.com/chef-cookbooks/audit#configure-node) and the errors and confusion that happens when users try to use hash syntax to add a profile to the default array format.  

One difficulty is the report handler pulls the list of profiles from the node attributes.  Which it appears cannot be modified inside a custom resource. To get around this, I've changed it to use `node.run_state`, which is transient and doesn't get saved to the chef-server.  If someone needs those to show up there, then they can still use the old method to do that.  

### Issues Resolved

### Check List

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -194,3 +194,13 @@ suites:
             url: https://github.com/dev-sec/tests-ssh-hardening/archive/master.zip
           ssh-baseline:
             supermarket: dev-sec/ssh-baseline
+  - name: custom_resource
+    run_list:
+      - recipe[audit::default]
+      - recipe[test_helper::custom_resources]
+    attributes:
+      audit:
+        reporter: json-file
+        profiles:
+          ssh-hardening:
+            url: https://github.com/dev-sec/tests-ssh-hardening/archive/master.zip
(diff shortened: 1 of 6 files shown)


btm:
@teknofire what was the issue you had with modifying the node attribute from inside the resource? 

Was it something like this?
```
Chef::Exceptions::ImmutableAttributeModification (Node attributes are read-only when you do not specify which precedence level to set. To set an attribute use code like `node.default["key"] = "value"')
```

lamont-granquist:
I was wondering if that was more of a compile-converge issue due to custom resource execution order (it will always be in the converge phase of the outer run context), but that would affect the node.run_state as well.

teknofire:
@btm @lamont-granquist There was no error, I didn't see the updated attributes in the audit report handler.  Although at this point I think I've figured out how to fix that now using the override precedence although I'm having to make a few more adjustments as it's a `VividMash` at that point and push additional profiles as an array like I was attempting.

teknofire:
Although then it gets into the issue of who should win when setting the attributes and making sure something doesn't come in later modifying the format I specified which is why I was leaning more towards using the run_state with its own set of key names to prevent any confusion on what should win based on attribute precedence.

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #358, opened 2019-02-13
Remove deprecated examples using Array of Hashes

jerryaldrichiii:
Due to https://github.com/chef-cookbooks/audit/pull/328 we must update the README.md to remove references to the Array of Hashes syntax.

```ruby
default['audit']['profiles'].push(
  name: 'example',
  compliance: 'admin/example'
)
```

becomes

```ruby
default['audit']['profiles'] = {}
default['audit']['profiles']['example'] = {
  compliance: 'admin/example'
}
```

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -85,6 +85,8 @@ default['audit']['overwrite'] = true
 
 # Chef Inspec Compliance profiles to be used for scan of node
 # See README.md for details
+# TODO: Make this `{}` in next major version
+# See: https://github.com/chef-cookbooks/audit/pull/328
 default['audit']['profiles'] = []
 
 # Attributes used to run the given profiles
--- a/README.md
+++ b/README.md
@@ -109,63 +109,43 @@ Note on AIX Support:
 
 ### Configure node
 
-Once the cookbook is available in Chef Server, you need to add the `audit::default` recipe to the run-list of each node. The profiles are selected using the `node['audit']['profiles']` attribute. A list of example configurations are documented in [Supported Configurations](docs/supported_configuration.md). Below are some other examples:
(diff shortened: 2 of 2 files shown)


attributes/default.rb
```diff
@@ -85,6 +85,8 @@
 
 # Chef Inspec Compliance profiles to be used for scan of node
 # See README.md for details
+# TODO: Make this `{}` in next major version
+# See: https://github.com/chef-cookbooks/audit/pull/328
 default['audit']['profiles'] = []
```
Vasu1105: I am not sure but removing examples will not resolve the main issue of hash of hashes #328  as we still having default as array.

jerryaldrichiii: Correct, I was not aware Hash of Hashes was broken. We are going to leave the current docs for now, then when we get Hash of Hashes working we will make sure we include setting the parent to `{}` in all examples.

teknofire: we'll also need to update the Automate 2 examples for when this is released as they use the `.push` method as well

jerryaldrichiii: @teknofire can you add any links to https://github.com/chef-cookbooks/audit/issues/359

Want to make sure we track them in all the places.

teknofire: https://automate.chef.io/docs/data-collection/#setting-up-chef-client-to-send-compliance-scan-data-through-the-chef-server-to-chef-automate

teknofire: https://github.com/chef/automate/blob/master/components/automate-chef-io/content/docs/data-collection.md

alexpop:
Merged as part of https://github.com/chef-cookbooks/audit/pull/386

Thanks!

PR #363, opened 2019-04-17
Export extra fields needed for rbac project tagging

alexpop:
Introducing a number of fields that are already expected by the `current` version of Automate.

Tested with:
* Node running in chef zero mode and reporting to Automate
* Policy file based node bootstraped with Managed Chef org
* Node bootstrapped with a Managed Chef org

I removed inspec 1 from the tests as it requires net-scp 1.2.1 which conflicts with net-scp 2.0.0 bundled in the latest chef-client.


files/default/handler/audit_report.rb
```diff
@@ -209,6 +209,13 @@ def gather_nodeinfo
           environment: n.environment,
           roles: runlist_roles,
           recipes: runlist_recipes,
+          policy_name: node.policy_name || '',
```
vjeffrey: why are some of these `node.blabla` and some `n.blabla` ?

libraries/helper.rb
```diff
@@ -71,6 +71,10 @@ def base_chef_server_url
     cs.to_s
   end
 
+  def chef_server_uri
+    URI(Chef::Config[:chef_server_url])
+  end
```
phiggins: Nitpick: if this key is ever not present for some reason, the error message could be improved by using `.fetch` instead of `[]`:

```
irb(main):001:0> require 'uri'
=> true
irb(main):002:0> h = {}
=> {}
irb(main):003:0> URI(h[:chef_server_url])
ArgumentError: bad argument (expected URI object or URI string)
	from /Users/pete/.asdf/installs/ruby/2.4.3/lib/ruby/2.4.0/uri/common.rb:739:in `URI'
	from (irb):3
	from /Users/pete/.asdf/installs/ruby/2.4.3/bin/irb:11:in `<main>'
irb(main):004:0> URI(h.fetch(:chef_server_url))
KeyError: key not found: :chef_server_url
	from (irb):4:in `fetch'
	from (irb):4
	from /Users/pete/.asdf/installs/ruby/2.4.3/bin/irb:11:in `<main>'
```

alexpop: I like it, thanks!

alexpop: Pete, looks like Chef::Config is a class and fetch is not working for it. But while I was at it, I found another way to do it without failing the run if chef_server_url is missing:

```ruby
[31] pry(#<Chef::Handler::AuditReport>)> {}.class
=> Hash
[32] pry(#<Chef::Handler::AuditReport>)> Chef::Config.class
=> Class
[33] pry(#<Chef::Handler::AuditReport>)> Chef::Config.fetch(:chef_server_url)
=> :chef_server_url
[34] pry(#<Chef::Handler::AuditReport>)> Chef::Config.chef_server_url
=> "chefzero://localhost:1"
[35] pry(#<Chef::Handler::AuditReport>)> URI(Chef::Config.missing_property || '')
=> #<URI::Generic >
[36] pry(#<Chef::Handler::AuditReport>)> URI(Chef::Config.missing_property || '').host || ''
=> ""
```

files/default/handler/audit_report.rb
alexpop: No good reason, `node` and `n` are the same. I pushed a new commits to use `n` in that function. Thank you

alexpop:
@phiggins and @vjeffrey I pushed a commit with changes based on your feedback. Thank you!

Issue #368, opened 2019-05-16
Cookbook broken with Chef-15


lamont-granquist:
See https://github.com/chef/chef/issues/8557

This mechanism has changed in Chef-15

https://github.com/chef-cookbooks/audit/blob/27df9721690ff9610d9fdee1074b9974dbe3163d/libraries/helper.rb#L17-L25

`node[:chef_guid]` is probably the best public API for that now.

lamont-granquist:
So the code should look more or less like this:

```
   # Returns the node's uuid
   def entity_uuid
    # the Chef::DataCollector::Messages API here is Chef < 15.0 backcompat and can be removed when Chef 14.x is no longer supported
    node[:chef_guid] || defined?(Chef::DataCollector::Messages) && Chef::DataCollector::Messages.node_uuid,
   end
```

The problem is that `node` may or may not be in the context.

It is also more than a bit weird that there appear to be spec tests for `it 'is not sending report when entity_uuid is missing' do` which are clearly not working correctly.

there's also lines in the audit reporter that look like this:

```
./files/default/handler/audit_report.rb:            entity_uuid: run_status.entity_uuid,
./files/default/handler/audit_report.rb:              entity_uuid: run_status.entity_uuid,
```

if that is the Chef::RunStatus object from core chef the it shouldn't have an entity_uuid method and didn't have on in Chef-14 so now I'm super confused.

zenspider:
_SINGLE SPACE INDENT?!?!?_ 😛

Thank you for changing it to `defined?(Chef::DataCollector::Messages)`.

lamont-granquist:
heh starchy copypasta error...

and yes...

alexpop:
@lamont-granquist thanks for assisting with this change.

I looked into it as well.
I don't see a problem with the `is not sending report when entity_uuid is missing` spec test. The `send_report` method is not sending the report if `entity_uuid` is missing, as I noticed in this output:

```
Running handlers:
[2019-05-31T09:01:53+00:00] WARN: Overwriting resource apache. To reference a specific version of apache use the resource() method
[2019-05-31T09:02:02+00:00] ERROR: entity_uuid() or run_id(dd3b05e8-b41f-43fa-a80b-c4d2c9385af7) can't be nil, not sending report to Chef Automate
  - Chef::Handler::AuditReport
```

`run_status.entity_uuid` in audit_report.rb is calling the `entity_uuid` helper method.
[...]

PR #371, opened 2019-05-31
Make entity_uuid work for Chef Infra 15

alexpop:
Fixes https://github.com/chef-cookbooks/audit/issues/368

### DIFF
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -229,6 +229,7 @@ class Chef
 
         # TODO: harmonize reporter interface
         if reporter == 'chef-visibility' || reporter == 'chef-automate'
+          # `run_status.entity_uuid` is calling the `entity_uuid` method in libraries/helper.rb
           opts = {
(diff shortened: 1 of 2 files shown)


lamont-granquist:
I dunno maybe it doesn't matter but this module gets included in a lot of different contexts where it looks like there my be no node object so that will blow up.  It seems to be used off of a `run_status` which seems to be the actual Chef::RunStatus object?  I don't understand how the wiring happens there.  It does inject itself into the Chef::Recipe DSL, but that shouldn't appear in the RunStatus object which inherits from nothing and AFAIK nothing gets mixed into it.

So....  this might work, but there still may be a pile of software developer footguns left lying around on the floor here...

This also clearly broke, but no tests went red.  That indicates that there really needs to be a test.

alexpop:
I would like to see a refactoring on how the node guid is identified.
For now, let's get this in to unblock Chef Infra 15 users. Travis tests are passing and I also manually tested it with the chef-automate reporter.

PR #374, opened 2019-06-20
Provide option to avoid saving the inspec attributes to the node object

alexpop:
Introducing a new cookbook attribute `['audit']['attributes_save']` to remove `['audit']['attributes']` from the node object if needed.

Thank you @trickyearlobe for the feedback!

Opening the PR for now while working on tests

### DIFF
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -41,7 +41,8 @@ class Chef
         end
         quiet = node['audit']['quiet']
         fetcher = node['audit']['fetcher']
-        attributes = node['audit']['attributes'].to_h
+
+        attributes = node.run_state['audit_attributes'].to_h
 
         # add chef node data as an attribute if enabled
         attributes['chef_node'] = chef_node_attribute_data if node['audit']['chef_node_attribute_enabled']
--- a/libraries/helper.rb
+++ b/libraries/helper.rb
@@ -116,6 +116,13 @@ module ReportHelpers
     File.expand_path('../../files/default/handler', __FILE__)
   end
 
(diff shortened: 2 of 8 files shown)


vjeffrey:
nice!

libraries/helper.rb
```diff
@@ -116,6 +116,13 @@ def cookbook_handler_path
     File.expand_path('../../files/default/handler', __FILE__)
   end
 
+  # Copies ['audit']['attributes'] into run_state for the audit_handler to read them later
+  # Deletes ['audit']['attributes'] if instructed by ['audit']['attributes_save']
+  def copy_audit_attributes
+    node.run_state['audit_attributes'] = node['audit']['attributes']
```
jerryaldrichiii: Mmmmm....yasss.

files/default/handler/audit_report.rb
```diff
@@ -41,7 +41,8 @@ def report
         end
         quiet = node['audit']['quiet']
         fetcher = node['audit']['fetcher']
-        attributes = node['audit']['attributes'].to_h
+
+        attributes = node.run_state['audit_attributes'].to_h
```
jerryaldrichiii: Going from a Mash to a Hash going to cause any issues? (I know it was a `.to_h` before, but I'm curious)

alexpop: Nobody reported issues with it and tests are showing the expected values.

alexpop:
Jerry and Victoria, thank you for the reviews!
I pushed another [commit](https://github.com/chef-cookbooks/audit/pull/374/commits/6dd9a3172bf681a511af284ace1ff82dc899192a) with ChefSpec tests to ensure further refactoring of the logic unexpectedly put the inspec attributes in the node object.

PR #375, opened 2019-06-20
Restore style, unit and chefspec testing

alexpop:
Adding back style, unit and chefspec testing in Travis after being accidentally removed in a previous PR.

### DIFF
--- a/resources/inspec_gem.rb
+++ b/resources/inspec_gem.rb
@@ -35,7 +35,7 @@ action :install do
       install_inspec_gem(version: new_resource.version, source: new_resource.source)
     end
   elsif new_resource.version.nil?
-    Chef::Log.info("inspec_gem: not installing Chef-InSpec. No Chef-Inspec version specified")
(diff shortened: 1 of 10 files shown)


resources/inspec_gem.rb
```diff
@@ -80,12 +80,12 @@ def compatible_with_client?(gem_version)
     return true if gem_version.nil?
 
     requirement = if chef_15?
-      # Chef-15 requires train 2.0 which was added in Inspec 4
-      Gem::Requirement.new('>= 4')
-    else
-      # min version required to run the audit handler
-      Gem::Requirement.new(['>= 1.25.1'])
-    end
+                    # Chef-15 requires train 2.0 which was added in Inspec 4
```
jerryaldrichiii: This seems super odd, maybe it's just the diff.

alexpop: I know! Linting wanted `if`, `else` and `end` aligned. That was my quick fix without refactoring the code.

PR #378, opened 2019-07-02
Ignore inspec_version for chef client >= 15

alexpop:
The `default['audit']['inspec_version']` attribute has been added to the cookbook before chef-client came with an embedded InSpec gem.

Currently the `inspec_gem` resource of the `audit` cookbook is uninstalling the existing versions of InSpec and installing the one specified by the `inspec_version` attribute. With the release of Chef Infra Client 15, this breaks `/opt/chef/bin/chef-client` as can be seen here:

```
[root@myapache ~]# chef-client
Traceback (most recent call last):
    3: from /usr/bin/chef-client:59:in `<main>'
    2: from /opt/chef/embedded/lib/ruby/2.6.0/rubygems/core_ext/kernel_gem.rb:65:in `gem'
    1: from /opt/chef/embedded/lib/ruby/2.6.0/rubygems/dependency.rb:323:in `to_spec'
/opt/chef/embedded/lib/ruby/2.6.0/rubygems/dependency.rb:313:in `to_specs': Could not find 'inspec-core' (= 4.3.2) - did find: [inspec-core-4.6.4]
```

I reproduced this issue by converging a node with `Chef Infra Client 15.0.300` and the `audit` cookbook with attribute `inspec_version` set to `4.6.4`. The `inspec_gem` resource uninstalled the embedded `inspec-core` version `4.3.2` and installed `4.6.4`.

With this PR the audit cookbook ignores the `inspec_version` attribute when Chef Infra Client version 15+ is detected and uses the embedded InSpec gem for the scan. Not breaking chef-client at this moment is more important than using a specific version of InSpec for the scan.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -5,7 +5,7 @@ maintainer_email 'cookbooks@chef.io'
 license 'Apache-2.0'
 description 'Allows for fetching and executing compliance profiles, and reporting its results'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version '7.8.0'
+version '8.0.0'
 
 source_url 'https://github.com/chef-cookbooks/audit'
 issues_url 'https://github.com/chef-cookbooks/audit/issues'
--- a/README.md
+++ b/README.md
@@ -117,6 +117,8 @@ Also beginning with version 3.x of the `audit` cookbook, the default version of 
 
 To install a different version of the InSpec gem, or to force installation of the gem, set the `node['audit']['inspec_version']` attribute to the version you wish to be installed.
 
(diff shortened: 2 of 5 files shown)


metadata.rb
```diff
@@ -5,7 +5,7 @@
 license 'Apache-2.0'
 description 'Allows for fetching and executing compliance profiles, and reporting its results'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version '7.8.0'
+version '8.0.0'
```
teknofire: Should update the README support matrix entry for Chef Client 15 to `>= 8.0.0` as well

alexpop: Yes, pushed a commit for that, thank you!

Issue #359, opened 2019-02-13
Remove `.push()` examples from documentation


jerryaldrichiii:
With the merging of https://github.com/chef-cookbooks/audit/pull/328 using an Array of Hashes for `node['audit']['profiles']` is deprecated. As such we should change all of the documentation to reflect this. This issue is a place to track this work.

Here are some of the sites I found:
  - [X] This repo's README.md (https://github.com/chef-cookbooks/audit/pull/358)
  - [ ] https://docs.chef.io/audit_cookbook.html (doesn't use .push but is still an Array of Hashes)
  - [ ] https://docs.chef.io/integrate_compliance_chef_server.html (doesn't use .push but is still an Array of Hashes)
  - [ ] https://automate.chef.io/docs/data-collection/
  - [ ] https://blog.chef.io/2016/11/09/the-audit-cookbook-a-how-to/

As a side note I feel we point all references on how to use the Audit cookbook to a single location.

@kagarmoe @mjingle thoughts?

schisamo:
@jerryaldrichiii It appears the new hash of hashes format doesn't actually work yet:
https://github.com/chef-cookbooks/audit/issues/339

Shouldn't we get this fixed before attempting to update documentation?

alexpop:
Done, closing. Thank you

PR #385, opened 2019-09-10
Change profiles default to a Hash

lamont-granquist:
This is a necessary breaking change.  Major version must be bumped after
merging this.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -1,7 +1,7 @@
 # encoding: utf-8
 #
 # Author:: Stephan Renatus <srenatus@chef.io>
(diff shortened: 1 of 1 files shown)


teknofire:
This is to fix #339 

There are two follow up PR's to clean up the examples and docs to only show Hash of Hash:  #358 and #359

lamont-granquist:
For future people who find this and would prefer backcompat you can always write a wrapper cookbook acmeco_audit which depends on the audit cookbook and restores the array, and then point your run_lists or depends lines at the wrapper instead.

```ruby
```

At some point that may break though.  That should be a temporary measure, and you should immediately start work on migrating to hashes.

Issue #403, opened 2019-12-16
Convert CI to buildkite


kekaichinose:
During all but one* Travis run, we see this error:

```
2.24s$ rake $SUITE
Problem loading gems please install chef and github_changelog_generator
sh -c kitchen test -c 1 chef15-compatible-inspec-ubuntu-1804
-----> Starting Kitchen (v2.3.3)
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::UserError
>>>>>> Message: Vagrant 1.1.0 or higher is not installed. Please download a package from https://www.vagrantup.com/downloads.html.
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

The one passing travis run is s a fluke, because it is looking for a failing outcome.

UPDATE: The vagrant error was spurious; it was simply looking at the wrong kitchenfile. #405 gets the tests passing again in Travis. 

It remains, however, to convert the CI suite over to Buildkite.

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #405, opened 2019-12-16
Get most test kitchen tests passing in Travis

### LINKED ISSUE #407: Apply current cookstyle
### Cookbook version
9.0.1

Currently has ~50 violations.

clintoncwolfe:
This was mainly about updating which Kitchenfile was to be used, and the fact that we now need to accept a license.

Along the way, one set of suites (5) was discovered to be badly broken due to an inspec bug (the naming collision around the `gem` resource). I opened an issue (#406) and commented out those suites for now.

~~Finally, the tests won't turn green yet because the codebase fails linting. I opened #407 to track that. I can fix that on this PR if desired.~~

Fixes #407, partially addresses #403

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,62 +28,59 @@ before_install:
 - CHEF_LICENSE="accept-no-persist" sudo chef gem install webmock
 - eval "$(chef shell-init bash)"
 - chef --version
-- cookstyle --version
-- foodcritic --version
-- kitchen --version
 - kitchen list
 
 matrix:
   include:
   # Run the style, unit and chefspec tests
-  - rvm: 2.6.3
+  - rvm: 2.6.5
     script: rake
   # Run integration tests with test-kitchen
-  - rvm: 2.6.3
+  - rvm: 2.6.5
     script: rake $SUITE
     env: SUITE=test:integration OS='default-centos-7'
-  - rvm: 2.6.3
+  - rvm: 2.6.5
     script: rake $SUITE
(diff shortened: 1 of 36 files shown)


libraries/matchers.rb
```diff
@@ -1,12 +1,3 @@
 # encoding: utf-8
 
 # used by ChefSpec
```
tas50: This whole file should just get removed now that it's empty

.travis.yml
```diff
@@ -17,13 +17,21 @@ branches:
...
+    - CHEF_VERSION=15
+    - KITCHEN_YAML=kitchen.dokken.yml
+    - CHEF_LICENSE="accept-no-persist"
+
 before_install:
 - sudo iptables -L DOCKER || ( echo "DOCKER iptables chain missing" ; sudo iptables -N DOCKER )
 - CHEF_LICENSE="accept-no-persist" sudo chef gem install webmock
 - eval "$(chef shell-init bash)"
 - chef --version
 - cookstyle --version
```
tas50: chef --version should include all the version output you need. It includes cookstyle and kitchen and foodcritic is dead.

.travis.yml
```diff
@@ -43,24 +51,27 @@ matrix:
   - rvm: 2.6.3
```
tas50: These should probably get updated to 2.6.5 while you're in here to match chef

alexpop:
Gents, I pused a commit that clears all the outstanding travis errors.
Once this merges, I want to get the other PRs rebased and merged as well. Cheers!

PR #417, opened 2020-04-15
Update warnings on size limits, error on nil uuid

btm:
- The default Infra Server limit is 2mb as of 13.0
- Add a failiure to the cs_automate reporter on a nil uuid that matches the automate reporter
- Cookstyle fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ addons:
     sources:
       - chef-current-xenial
     packages:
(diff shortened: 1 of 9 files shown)


alexpop:
Test passed after pinning CHEF_VERSION to `15.8.23`
As @teknofire pointed out to me, something is fishy with the bundle pry gems in chef `15.9.17`

PR #418, opened 2020-04-16
Support the Automate /compliance/profiles/metasearch endpoint to send reduced reports

alexpop:
### Description

Automate has recently introduced a new API to check which profiles ids are missing in the database. This API enables the audit cookbook to avoid sending to Automate full reports that can contain up to 90% profile meta information that Automate already has.

### Issues Resolved

The Automate(grpc) report limit of 4MB will be harder to reach with this change.


attributes/default.rb
```diff
@@ -88,3 +88,7 @@
 # <chef_cache_path>/cookbooks/audit/inspec-<YYYYMMDDHHMMSS>.json
 default['audit']['json_file']['location'] =
   File.expand_path(Time.now.utc.strftime('../../inspec-%Y%m%d%H%M%S.json'), __FILE__)
+
+# Controls that have a run_time below this limit will be stripped of the `start_time` and
+# `start_time` fields to reduce the size of the reports being sent to Automate
```
vjeffrey, line 93: nitpick: did you mean start time and end time?

vjeffrey: does this mean the report wouldnt have an end time? is that ok with automate ingestion?

alexpop: These are for the control results. Let me rephrase the comment to be more clear about this:

```
# Control results that have a `run_time` below this limit will be stripped of the 
# `start_time` and `run_time` fields to reduce the size of the reports being sent to Automate
```

Issue #413, opened 2020-03-19
missing waiver file fails all reporting


mattray:
If you pass a waiver file via an attribute
```
node['audit']['waiver_file'] = '/root/waivers.yml'
```
and that file is not there, InSpec throws an exception and no reports are sent to Automate. This makes it impossible to have a placeholder for dropping waivers on a filesystem, if it's not there you'll get no reporting at all.

Exception
```
[2020-03-19T17:42:59+11:00] ERROR: Report handler Chef::Handler::AuditReport raised #<Inspec::Exceptions::InputsFileDoesNotExist: Cannot find input file '/root/waivers.yml'. Check to make sure file exists.>
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/input_registry.rb:232:in `validate_inputs_file_readability!'
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/input_registry.rb:205:in `block in bind_inputs_from_input_files'
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/input_registry.rb:204:in `each'
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/input_registry.rb:204:in `bind_inputs_from_input_files'
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/input_registry.rb:145:in `bind_profile_inputs'
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/input_registry.rb:311:in `block (2 levels) in <class:InputRegistry>'
[2020-03-19T17:42:59+11:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-4.18.85/lib/inspec/profile.rb:134:in `initialize'
[...]

PR #420, opened 2020-04-21
Ignore missing waiver files

alexpop:
Fixes https://github.com/chef-cookbooks/audit/issues/413


alexpop:
Thank you for the reviews, merging and releasing v7.2.1 with this fix in

PR #430, opened 2020-07-09
Report even on inspec runtime exceptions

alexpop:
### Description

Before this, InSpec exceptions caused by invalid profiles or invalid InSpec attributes that didn't produce a valid report were not being sent over to Automate / Chef-Server.

### Issues Resolved

Catches any InSpec runtime exception and generates a minimal failed json report.

### DIFF
--- a/files/default/handler/audit_report.rb
+++ b/files/default/handler/audit_report.rb
@@ -198,13 +198,33 @@ class Chef
           Chef::Log.debug "Audit Report #{r}"
           r
         else
-          Chef::Log.warn 'No audit tests are defined.'
-          {}
+          failed_report('No audit tests are defined.')
         end
       rescue Inspec::FetcherFailure => e
-        Chef::Log.error e.message
-        Chef::Log.error "We cannot fetch all profiles: #{tests}. Please make sure you're authenticated and the server is reachable."
-        {}
+        err = "Cannot fetch all profiles: #{tests}. Please make sure you're authenticated and the server is reachable. #{e.message}"
+        failed_report(err)
+      rescue => e
+        failed_report(e.message)
+      end
+
+      # In case InSpec raises a runtime exception without providing a valid report,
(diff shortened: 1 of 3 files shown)


files/default/handler/audit_report.rb
```diff
@@ -198,13 +198,33 @@ def call(opts, profiles)
...
-          {}
+          failed_report('No audit tests are defined.')
         end
       rescue Inspec::FetcherFailure => e
-        Chef::Log.error e.message
-        Chef::Log.error "We cannot fetch all profiles: #{tests}. Please make sure you're authenticated and the server is reachable."
-        {}
+        err = "Cannot fetch all profiles: #{tests}. Please make sure you're authenticated and the server is reachable. #{e.message}"
+        failed_report(err)
+      rescue => e
```
Schwad, line 206: Thanks for doing this! 🙏 

I know we have put in recent work to ensure that we articulate failed profiles now instead of crashing or raising an error, particularly with: https://github.com/inspec/inspec/pull/5128 (which came out in `4.22`) 

To test it this could cover instances where we simply have ruby code that cannot be evaluated on the profile level before running tests, i.e.:

```ruby
#my_profile.rb

fail('This profile is broken!')

control 'my-control' do
  # code
end
```

And produce:

`"ERROR: Failed to load profile raise_outside_control: Failed to load source for controls/my-profile.rb: This profile is broken!"`

Which will be included in the actual report generated instead of a ruby-level raise.

Looking at the tests and description, I'm assuming this is meant to catch edge cases where config is incorrect, profile is nonexistent and just be a general catch-all of other unforeseen failures?

If so, looks good to me but wanted to make sure I had clarity on the case here.

alexpop: Yes, that is correct. Any other error that InSpec is not able to report in the json format will be captured and reported with the minimal json report template in this PR.

PR #450, opened 2021-01-26
Add CLI reporter

aknarts:
### Description

Adds a CLI reporter for output to chef log(similar to old Audit phase report)
Screenshot from 2021-01-26 19-07-27

### Check List

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -28,7 +28,7 @@ default['audit']['inspec_gem_source'] = nil
 default['audit']['inspec_backend_cache'] = true
 
 # controls where inspec scan reports are sent
-# possible values: 'chef-server-automate', 'chef-automate', 'json-file'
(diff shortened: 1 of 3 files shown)


tas50:
@aknarts Thanks for taking the time to put this PR together, but we've actually integrated the code from the audit cookbook directly into Chef Infra Client as part of a new Compliance Phase. The codebase is mostly identical to this cookbook so you could probably just add this there. The goal is to deprecate this cookbook so that users can just run compliance scans on their nodes without an additional dependency.

aknarts:
> @aknarts Thanks for taking the time to put this PR together, but we've actually integrated the code from the audit cookbook directly into Chef Infra Client as part of a new Compliance Phase.

I was not aware of this. I am in the process of migrating to the latest Chef Infra Client(16), what version is this supposed to be in/are there any docs for it?

aknarts:
Found the PR https://github.com/chef/chef/pull/10547 closing in favour of that

